PR #1, opened 2018-02-24
Add metadata_props and descriptions to CoreML converter

lutzroeder:


onnxmltools/convert/coreml/convert.py
```diff
@@ -84,7 +84,20 @@ def convert(model, name=None, doc_string=''):
...
 
-    mb = ModelBuilder(name, doc_string)
+    # Convert CoreML description, author and license
+    metadata = spec.description.metadata
+    metadata_props = None
+    if metadata:
+        metadata_props = []
+        if metadata.shortDescription:
+            if not doc_string:
+                doc_string = metadata.shortDescription
```
wschin: It means sometimes we add shortDescription and sometime we don't. For consistency, can we do something like creating a StringStringEntryProto to store this value (key: 'short description from CoreML model', value: metadata.shortDescription')?

onnxmltools/convert/coreml/datatype.py
```diff
@@ -42,31 +42,31 @@ def _convert(cm_type):
...
     key_type = cm_value.type.dictionaryType.WhichOneof('KeyType')
     onnx_key_type = _convert(key_type)
     onnx_value_type = onnx_proto.TensorProto.FLOAT
-    map_type = model_util.make_map_value_info(cm_value.name, onnx_key_type, onnx_value_type)
+    map_type = model_util.make_map_value_info(cm_value.name, onnx_key_type, onnx_value_type, doc_string)
     return map_type
 
 
-def _handle_image_feature(cm_value, batch_size=1):
+def _handle_image_feature(cm_value, batch_size=1, doc_string=''):
```
wschin: This function should not include doc_string as its argument since doc_string has been used to describe the tensor format of image type. Another possibility is to concatenate our string and the short description from CoreML.

onnxmltools/convert/coreml/datatype.py
```diff
@@ -96,14 +96,18 @@ def make_value_info(value, batch_size=1):
     except Exception:
         raise Exception("Not a valid CoreML type")
 
+    doc_string = ''
```
wschin: I think we can directly use value.shortDescription to initialize doc_string because we can get default value as least.

onnxmltools/convert/coreml/convert.py
lutzroeder: It means `shortDescription` is always used unless the user explicitly overrides it when calling `convert`. `shortDescription` should be shown in tools the same way as `doc_string`, adding another description property seems arbitrary. A better approach would be to discourage passing `doc_string` to `convert` as the CoreML file should already contain the description.

onnxmltools/convert/coreml/datatype.py
lutzroeder: Adding concatenate.

onnxmltools/convert/coreml/convert.py
wschin: Sounds good. I am fine with this change now.

lutzroeder: Done.

PR #4, opened 2018-02-27
Specify opset_import

wschin:
ModelProto (https://github.com/onnx/onnx/blob/b0ffb2d302b3da405ceeeee29680e3e313101b1a/onnx/onnx-ml.proto3#L176) needs opset_import to pass ONNXChecker.


onnxmltools/convert/common/model_util.py
```diff
@@ -90,6 +90,14 @@ def make_model(name, ir_version, producer, producer_version, domain, model_versi
     model.domain = domain
     model.model_version = model_version
     model.doc_string = doc_string
+    # Neural network operators' version
+    opset = model.opset_import.add()
+    opset.domain = ''
+    opset.version = 3
+    # Traditional ML operators' version
```
prasanthpul: This should be conditional and added only to models that actually use ops from that namespace.

wschin: Thanks. Now we collect the operator set information from existing operators only.

onnxmltools/convert/common/model_util.py
```diff
@@ -14,6 +14,137 @@
 onnx_integer_types = [onnx_proto.TensorProto.UINT8, onnx_proto.TensorProto.INT8, onnx_proto.TensorProto.UINT16,
                       onnx_proto.TensorProto.INT16, onnx_proto.TensorProto.INT32, onnx_proto.TensorProto.INT64]
 
+onnx_operator_version_info = {
```
singlis: There is another table similar to this under make_node that contains a list of operators that should be set to ai.onnx.ml for the domain name. Is there anyway that the make_node code could use this version info structure instead? This would reduce duplication.

wschin: You're right. Let me try to move this into converters and make things cleaner.

PR #9, opened 2018-03-05
Handling of model version information

singlis:
- Moving model version, domain, producer, and producer version into functions. This allows for other packages that are using onnxmltools to override these variables as needed.


onnxmltools/convert/common/model_util.py
```diff
@@ -9,12 +9,44 @@
...
 
+
+
 # Includes basic ONNX data types only, advanced data types will be supported later.
 # This list is used to convert integers to floats.
 onnx_integer_types = [onnx_proto.TensorProto.UINT8, onnx_proto.TensorProto.INT8, onnx_proto.TensorProto.UINT16,
                       onnx_proto.TensorProto.INT16, onnx_proto.TensorProto.INT32, onnx_proto.TensorProto.INT64]
 
 
+def producer():
```
wschin: Would it be better to rename producer to get_producer?

onnxmltools/convert/common/model_util.py
```diff
@@ -9,12 +9,44 @@
...
+    """
+    Internal helper function to return the producer version
+    """
+    from ... import __producer_version__
+    return __producer_version__
+
+def domain():
+    """
+    Internal helper function to return the model domain
+    :return:
```
wschin: Maybe you want to remove "return"

onnxmltools/convert/common/model_util.py
```diff
@@ -9,12 +9,44 @@
...
                       onnx_proto.TensorProto.INT16, onnx_proto.TensorProto.INT32, onnx_proto.TensorProto.INT64]
 
 
+def producer():
+    """
+    Internal helper function to return the producer
+    """
+    from ... import __producer__
+    return __producer__
+
```
wschin: Please add one more empty line between two consecutive functions.

onnxmltools/convert/common/model_util.py
```diff
@@ -9,12 +9,44 @@
 from ...proto import onnx_proto
 from ...proto import helper
 
+
+
```
wschin: Too many empty lines.

onnxmltools/convert/common/model_util.py
singlis: I can go either way, but I see your point - I think the get term makes it sound that its retrieving rather than doing something. I will update.

onnxmltools/convert/common/model_util.py
singlis: will do

Issue #16, opened 2018-04-25
pip from source, then import prompt ImportError: cannot import name 'onnx_ml_pb2'


FrozenGene:
After pip from source and import

>>> import onnxmltools
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/blue/Documents/third-party/onnxmltools/onnxmltools/__init__.py", line 23, in <module>
    from .utils import load_model
  File "/Users/blue/Documents/third-party/onnxmltools/onnxmltools/utils/__init__.py", line 7, in <module>
    from .main import load_model
  File "/Users/blue/Documents/third-party/onnxmltools/onnxmltools/utils/main.py", line 8, in <module>
    from ..proto import onnx_proto
  File "/Users/blue/Documents/third-party/onnxmltools/onnxmltools/proto/__init__.py", line 16, in <module>
    from onnx import onnx_ml_pb2 as onnx_proto
ImportError: cannot import name 'onnx_ml_pb2'

However, pip install onnxmltools (i.e. from PyPI) doesn't have problem

FrozenGene:
Ok, I find that in the proto/__init__.py, we use from onnx import onnx_ml_pb2 as onnx_proto, however, we don't have onnx_ml_pb2 any more. Like TensorProto has been into the top level onnx, not onnx.onnx_ml_pb2. I have created one pull request https://github.com/onnx/onnxmltools/pull/17 to solve this problem.

wschin:
I am not following. To get `onnx-ml.proto` compiled by protobuf complier and produce `onnx_ml_pb2.py`, we need to specify `set ONNX_ML=1` (assuming you're using Windows command line) before installing `onnx`. One reason that we cannot use onnx.proto is that Core ML classifiers usually produce `Map `output and such a type is only defined in `onnx-ml.proto
`
.

FrozenGene:
@wschin Thanks for replying. I have set the environment of ONNX_ML=1 before installing onnx. It works. However, I think we should add the key information in our README.md about installing from source, otherwise, how do I know I should set this environment variables?

wschin:
Agree. Would you mind to create a PR?

FrozenGene:
@wschin Yes, I would create a PR.

PR #17, opened 2018-04-25
Fix the problem of inputting from source

FrozenGene:
See the detail of https://github.com/onnx/onnxmltools/issues/16


FrozenGene:
We can set the environment variable ONNX_ML=1 before installing onnx to work.

PR #35, opened 2018-05-11
Fix list index out of range

weikexin:


wschin:
@shauheen , this fix looks good to me and it passed the corresponding end-to-end tests. Can we merge?

wschin:
@Shauheen, ping again. Thanks.

PR #46, opened 2018-05-17
Add float16 support

duli2012:
We don't need this PR since we plan to produce a stand-along float16 converter.

### DIFF
--- a/onnxmltools/convert/common/_container.py
+++ b/onnxmltools/convert/common/_container.py
@@ -113,7 +113,7 @@ class ModelComponentContainer:
     encapsulated in a ONNX ModelProto.
     '''
 
-    def __init__(self, float_precision=32, targeted_onnx):
+    def __init__(self, targeted_onnx, float_precision=32):
         '''
         :param float_precision: precision of float number, 32 bits or 16 bits
         :param targeted_onnx: A string, for example, '1.1.2' and '1.2'.
--- a/onnxmltools/convert/common/_container.py
+++ b/onnxmltools/convert/common/_container.py
@@ -6,6 +6,8 @@
 
 import six
 from ...proto import helper, onnx_proto
+from distutils.version import StrictVersion
+from ...proto import helper
 
 
 class RawModelContainer(object):
@@ -111,7 +113,11 @@ class ModelComponentContainer:
     encapsulated in a ONNX ModelProto.
     '''
 
-    def __init__(self, float_precision=32):
+    def __init__(self, float_precision=32, targeted_onnx):
+        '''
+        :param float_precision: precision of float number, 32 bits or 16 bits
+        :param targeted_onnx: A string, for example, '1.1.2' and '1.2'.
+        '''
         # Inputs of ONNX graph. They are ValueInfoProto in ONNX.
         self.inputs = []
         # Outputs of ONNX graph. They are ValueInfoProto in ONNX.
@@ -126,6 +132,8 @@ class ModelComponentContainer:
         self.node_domain_version_pair_sets = set()
         # Precision of float numbers: 16 or 32
         self.float_precision = float_precision
+        # The targeted ONNX version. All produced operators should be supported by the targeted ONNX version.
+        self.targeted_onnx_version = StrictVersion(targeted_onnx)
 
     def _make_value_info(self, variable):
         value_info = helper.ValueInfoProto()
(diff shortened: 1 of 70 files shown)


onnxmltools/convert/common/_container.py
```diff
@@ -128,13 +130,18 @@ def __init__(self, targeted_onnx):
         self.nodes = []
         # ONNX operators' domain-version pair set. They will be added into opset_import field in the final ONNX model.
         self.node_domain_version_pair_sets = set()
+        # Precision of float numbers: 16 or 32
```
wschin: Please throw an error if it's neither 16 or 32. Also, do we need to support Double?

onnxmltools/convert/common/_container.py
```diff
@@ -112,8 +113,9 @@ class ModelComponentContainer:
     encapsulated in a ONNX ModelProto.
     '''
 
-    def __init__(self, targeted_onnx):
+    def __init__(self, float_precision=32, targeted_onnx):
         '''
+        :param float_precision: precision of float number, 32 bits or 16 bits
```
wschin: Precision of floats (or Precision of floating-point numbers), 32 bits or 16 bits. Applicable to ONNX inputs, outputs and initializers.

onnxmltools/convert/common/_container.py
```diff
@@ -112,8 +113,9 @@ class ModelComponentContainer:
     encapsulated in a ONNX ModelProto.
     '''
 
-    def __init__(self, targeted_onnx):
+    def __init__(self, float_precision=32, targeted_onnx):
```
wschin: this signature should be (self, targeted_onnx, float_precision=32)

onnxmltools/convert/common/_container.py
```diff
@@ -164,6 +171,8 @@ def add_initializer(self, name, onnx_type, shape, content):
         :param shape: Tensor shape, a list of integers.
         :param content: Flattened tensor values (i.e., a float list or a float array).
         '''
+        if (self.float_precision == 16 and onnx_type == onnx_proto.TensorProto.FLOAT):
```
wschin: Please remove "(" and ")".

onnxmltools/convert/common/_container.py
```diff
@@ -5,6 +5,7 @@
 # --------------------------------------------------------------------------
 
 import six
+from ...proto import helper, onnx_proto
```
wschin: from distutils.version import StrictVersion 
from ...proto import helper, onnx_proto

wschin:
Please also convert floating-point numbers into 16 bits if a Node have float attributes. This can be done in ModelComponentContainer's add_node(…).

onnxmltools/convert/common/_container.py
duli2012: done

onnxmltools/convert/common/_container.py
duli2012: done

onnxmltools/convert/common/_container.py
duli2012: done

wschin:
Also add tests please. Thanks!

justinormont:
Will we support bfloat16?

https://en.wikipedia.org/wiki/Bfloat16_floating-point_format

duli2012:
We'll support IEEE float16 in another standalone converter.

justinormont:
@duli2012: [bfloat16](https://en.wikipedia.org/wiki/Bfloat16_floating-point_format) differs from [IEEE float16](https://en.wikipedia.org/wiki/Half-precision_floating-point_format). 

The bfloat16 is a truncated float32 which maintains the dynamic range of the float32 (~1e-38 to ~3e38.) while reducing the significand's precision. 

The gain is to avoid the overflow/underflow that a standard float16 would hit due to its limited range (~6e-8 to 65504). In a float16, an exploding gradient can very easily hit the 66k limit of a float16. This otherwise is rather tricky for a normal coder to avoid in the float16, but easily handled in the bfloat16's larger range. Tensorflow's ops normally use bfloat16 in mixed precision by doing multiplies in bfloat16 and summation into float32.  

Conversion to/from float32 to bfloat16 is quite trivial as seen in Tensorflow's [bfloat16 code](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/bfloat16.cc).

The main limiting factor for why to not support bfloat16 in ONNX is the lack of current hardware support. Only Google's TPUs, and various upcoming Intel processors support bfloat16, which limits the number of models which will contain bfloat16. Whereas IEEE float16 has wide support in GPUs.

duli2012:
Thanks for explanation, @justinormont .
Our current plan is suppport IEEE float16.

PR #73, opened 2018-06-02
Add two tests

duli2012:


tests/end2end/test_single_operator_with_cntk_backend.py
```diff
@@ -172,6 +173,12 @@ def test_dense(self):
 
         self.assertTrue(np.allclose(y_reference, y_produced))
 
+        # Verify Keras-to-ONNX path
+        onnx_model = onnxmltools.convert_keras(keras_model)
+        y_produced = _evaluate(onnx_model, x).reshape(N, D)
```
wschin: Please add a note to explain the usage of reshape(N, D). E.g., the output shape, [1, N, D], from CNTK got an extra 1-element dimension, so we squeeze it with reshape. Please also copy this note to other places where Reshape is used.

tests/end2end/test_single_operator_with_cntk_backend.py
```diff
@@ -392,13 +399,20 @@ def test_recursive_model(self):
         mapped2_2 = sub_model2(input2)
         sub_sum = Add()([mapped1_2, mapped2_2])
         keras_model = Model(inputs=[input1, input2], output=sub_sum)
+        x = [x, 2*x]
```
wschin: Please move it right below Line 386.

PR #76, opened 2018-06-05
Float16 converter

stevenlix:


stevenlix:
Sure. I will add a unit test for it.

onnxmltools/utils/float16_converter.py
```diff
@@ -0,0 +1,110 @@
...
+    queue = []
+    if not model.graph: return None
+    queue.append(model)
+    while queue:
+        next_level = []
+        for q in queue:
+            # if q is model, push q.graph
+            try: q.graph
+            except: pass
+            else: next_level.append(q.graph)
```
duli2012: why not just if_else ?

stevenlix: because q stored in the queue could be model, graph, or attributes. q.graph may not exist if q is not model.

stevenlix:
replaced try exception with hasattr(), and convert raw_data attribute to float16 type

onnxmltools/utils/float16_converter.py
```diff
@@ -13,21 +13,23 @@ def npfloat16_to_int(np_list):
...
 
-def float16_converter(model_path):
+def float16_converter(onnx_file, new_onnx_file):
     """
-    Convert TensorProto type float to float16 in ONNX model.
-    :param model_path: ONNX model path and name
-    :return model: converted ONNX model
+    Convert ONNX model from TensorProto type float to float16.
+    :param onnx_file: ONNX model path and name
+    :param new_onnx_file: new ONNX model path and name
```
wschin: What does "and name" mean? Is this parameter a tuple of strings?

onnxmltools/utils/float16_converter.py
```diff
@@ -13,21 +13,23 @@ def npfloat16_to_int(np_list):
...
 
     Example:
 
     ::
 
         from onnxmltools.utils import float16_converter
-        model_path = 'c:/winmlperf_coreml_SqueezeNet_prerelease.onnx'
-        model = float16_converter(model_path)
+        onnx_file = 'c:/winmlperf_coreml_SqueezeNet_prerelease.onnx'
```
wschin: Do we have a better example name? Maybe change `c:/winmlperf_coreml_SqueezeNet_prerelease.onnx` to `example.onnx` and `c:/winmlperf_coreml_SqueezeNet_prerelease_float16.onnx` to `example_float16.onnx`?

onnxmltools/utils/float16_converter.py
```diff
@@ -13,21 +13,23 @@ def npfloat16_to_int(np_list):
     return [int(bin(_.view('H'))[2:].zfill(16), 2) for _ in np_list]
 
 
-def float16_converter(model_path):
+def float16_converter(onnx_file, new_onnx_file):
```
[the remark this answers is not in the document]
stevenlix: Yes. Here is an example,
        onnx_file = 'c:/winmlperf_coreml_SqueezeNet_prerelease.onnx'
        new_onnx_file = 'c:/winmlperf_coreml_SqueezeNet_prerelease_float16.onnx'
        float16_converter(onnx_file, new_onnx_file)

onnxmltools/utils/float16_converter.py
```diff
@@ -50,20 +52,20 @@ def float16_converter(model_path):
             if hasattr(q, 'graphs'):
                 for n in q.graphs:
                     next_level.append(n)
-            # graph.initializer  (TensorProto)
+            # if q is graph, process graph.initializer(TensorProto), input and output (ValueInfoProto)
```
wschin: Do we also need to handle intermediate variables (ValueInfoProto's which are neither inputs nor outputs)?

onnxmltools/utils/float16_converter.py
```diff
@@ -0,0 +1,118 @@
+import numpy as np
+from struct import unpack
+from ..convert.common.data_types import *
+from onnxmltools.utils import load_model, save_model
+
+
+def npfloat16_to_int(np_list):
```
wschin: Please change `npfloat16_to_int` to `_npfloat16_to_int` if it's an internal function. I know often I also forgot doing so but let's keep this as restrictive as possible. Thanks a lot.

onnxmltools/utils/float16_converter.py
```diff
@@ -0,0 +1,118 @@
...
+    Convert numpy float16 to python int.
+    :param np_list: numpy float16 list
+    :return int_list: python int list
+    """
+    return [int(bin(_.view('H'))[2:].zfill(16), 2) for _ in np_list]
+
+
+def float16_converter(onnx_file, new_onnx_file):
+    """
+    Convert ONNX model from TensorProto type float to float16.
```
wschin: Please add an empty line between function description and its parameters.

onnxmltools/utils/float16_converter.py
```diff
@@ -0,0 +1,118 @@
...
+def npfloat16_to_int(np_list):
+    """
+    Convert numpy float16 to python int.
+    :param np_list: numpy float16 list
+    :return int_list: python int list
+    """
+    return [int(bin(_.view('H'))[2:].zfill(16), 2) for _ in np_list]
+
+
+def float16_converter(onnx_file, new_onnx_file):
```
wschin: Does this function only convert FLOAT to FLOAT16? Or it also converts DOUBLE to FLOAT16?

onnxmltools/utils/float16_converter.py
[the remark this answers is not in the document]
wschin: I see. Do you think `def float16_converter(onnx_model)` a better interface? Using files as function I/O may add some friction when user has already loaded the model.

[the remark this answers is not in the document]
stevenlix: You mean use protobuf model rather than onnx file?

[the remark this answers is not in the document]
wschin: Yes!

onnxmltools/utils/float16_converter.py
wschin: If it only converts FLOAT to FLOAT16, we need to throw if a DOUBLE is found.

onnxmltools/utils/float16_converter.py
wschin: Couldn't we use if-else statement with `isinstance(q, ModelProto)`?

onnxmltools/utils/float16_converter.py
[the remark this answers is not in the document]
stevenlix: The intention of this converter is to serve as a standalone feature.  That's why it consumes/generates onnx file. But it can be easily done if we want it to consume/generate protobuf model

onnxmltools/utils/float16_converter.py
wschin: float16_converter sounds like a class, not a function. Maybe call this function `compress_model_float_to_float16`? I am not good at naming so you probably can find a better name.

onnxmltools/utils/float16_converter.py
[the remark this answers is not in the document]
stevenlix: ok. I can change the input/output to onnx model. If there is any need to save/read model to file, a separate call to save_model()/load_model() can do

onnxmltools/utils/float16_converter.py
stevenlix: I thought DOUBLE is a seperate data type to handle if we want to convert it. Same as complex64 or complex128. Do we need to consider them during the conversion from FLOAT to FLOAT16?

onnxmltools/utils/float16_converter.py
stevenlix: will do

onnxmltools/utils/float16_converter.py
stevenlix: will do

onnxmltools/utils/float16_converter.py
wschin: ONNX potentially allows a mixture of numerical types in a single model. Assume a user calls this float compressor and her model contains both of DOUBLE and FLOAT. It's a bit confusing if she can still find DOUBLE in her compressed model. This comment is relatively minor so it is not blocking you from merging this PR. :)

wschin:
Instead of checking in a binary model file (model will be updated frequently so it won't take too long to get a large repo while Git accumulates all historical objects), I'd prefer to create a two-operator models from scratch. For example, we can create one Conv (for testing initializer and input) + HardSigmoid (for testing attributes). Then, we can compare the values created before and after calling this converter.

onnxmltools/utils/float16_converter.py
```diff
@@ -0,0 +1,113 @@
+import numpy as np
+from struct import unpack
+from ..convert.common.data_types import *
+
+
+def _npfloat16_to_int(np_list):
+    """
```
wschin: Could you change `" `to `'`? We use `'` throughout the code base. Thanks.

onnxmltools/utils/float16_converter.py
```diff
@@ -0,0 +1,113 @@
...
+
+    :param np_list: numpy float16 list
+    :return int_list: python int list
+    """
+    return [int(bin(_.view('H'))[2:].zfill(16), 2) for _ in np_list]
+
+
+def float16_converter(model):
+    """
+    Convert ONNX model from TensorProto type float to float16.
```
wschin: Maybe you can use this help message: Convert floats in the specified ONNX model to float16.

onnxmltools/utils/float16_converter.py
```diff
@@ -0,0 +1,113 @@
...
+
+    :param model: ONNX model
+    :return: converted ONNX model
+
+    Example:
+
+    ::
+
+        from onnxmltools.utils import float16_converter
+        new_onnx_model = float16_converter(onnx_model)
```
wschin: If you change the function name, please adjust the example here accordingly.

onnxmltools/utils/float16_converter.py
```diff
@@ -0,0 +1,113 @@
...
+                    if hasattr(n, 'type'):
+                        if n.type.tensor_type.elem_type == onnx_proto.TensorProto.FLOAT:
+                            n.type.tensor_type.elem_type = onnx_proto.TensorProto.FLOAT16
+            # if q is graph, process graph.value_info (ValueInfoProto)
+            if hasattr(q, 'value_info'):
+                for n in q.value_info:
+                    if n.type.tensor_type.elem_type == onnx_proto.TensorProto.FLOAT:
+                        n.type.tensor_type.elem_type = onnx_proto.TensorProto.FLOAT16
+            # if q is node.attribute, process node.attribute.t and node.attribute.tensors (TensorProto)
+            if hasattr(q, 't'):
```
wschin: It's a bit dangerous. The field `t` is not a very meaningful name so adding type checking before accessing attributes might be necessary.

onnxmltools/utils/float16_converter.py
```diff
@@ -0,0 +1,113 @@
...
+                if q.t.data_type == onnx_proto.TensorProto.FLOAT:
+                    q.t.data_type = onnx_proto.TensorProto.FLOAT16
+                    int_list = _npfloat16_to_int(np.float16(q.t.float_data))
+                    q.t.int32_data[:] = int_list
+                    q.t.float_data[:] = []
+                    # convert q.t.raw_data (bytes) to float
+                    float32_list = []
```
wschin: If `float_data` is set, the field `raw_data` must be empty so we should skip the subsequent lines. To determine which filed is set, please see [this](https://github.com/onnx/onnx/blob/4b4085c2e9d5a944651a2dd0dfdd20ef452bdcdf/onnx/onnx.in.proto#L319).

onnxmltools/utils/float16_converter.py
```diff
@@ -0,0 +1,113 @@
...
+                    for i in range(len(q.t.raw_data) // 4):
+                        float32_list += unpack('f', q.t.raw_data[i * 4:(i + 1) * 4])
+                    # convert float to float16
+                    int_list = _npfloat16_to_int(np.float16(float32_list))
+                    # convert float16 to bytes and write back to raw_data
+                    q.t.raw_data = b''
+                    for num in int_list:
+                        q.t.raw_data += num.to_bytes(2, byteorder='little', signed=False)
```
wschin: Cool!

onnxmltools/utils/float16_converter.py
```diff
@@ -0,0 +1,113 @@
...
+            if hasattr(q, 'tensors'):
```
wschin: The name `tensors` is not very informative. Maybe adding a type checking later. See my comment for `t`.

tests/utils/test_utils.py
```diff
@@ -69,3 +69,10 @@ def test_set_docstring_blank(self):
         self.assertRaises(ValueError, set_model_doc_string, onnx_model.doc_string, "sample")
         set_model_doc_string(onnx_model, "", True)
         self.assertEqual(onnx_model.doc_string, "")
+
+    def test_float16_converter(self):
```
wschin: Given that the binary file will be removed, please delete this test for now.

wschin:
You may also need to sign CLA.

PR #97, opened 2018-06-22
add copyright info and check input validation

stevenlix:


onnxmltools/utils/float16_converter.py
```diff
@@ -17,16 +22,28 @@ def convert_float_to_float16(model):
...
+    :return: converted ONNX model ProtoBuf object
 
     Example:
 
     ::
 
+        Example 1: Convert ONNX model ProtoBuf object:
+        from onnxmltools.utils.float16_converter import convert_float_to_float16
+        new_onnx_model = convert_float_to_float16(onnx_model)
+
```
wenbingl: This method is very intuitive, shall we just keep one example?

stevenlix: The thinking is that for external user they may want to know how to use this method as a standalone feature to convert onnx binary file.

onnxmltools/utils/float16_converter.py
```diff
@@ -17,16 +22,28 @@ def convert_float_to_float16(model):
...
         from onnxmltools.utils.float16_converter import convert_float_to_float16
+        from onnxmltools.utils import load_model, save_model
+        onnx_model = load_model("model.onnx")
         new_onnx_model = convert_float_to_float16(onnx_model)
+        save_model(new_onnx_model, "new_model.onnx")
+
     '''
+    # raise error if input is empty or not ModelProto object
+    if model is None or not isinstance(model, onnx_proto.ModelProto):
+        raise ValueError("model is not an onnx model")
```
wschin: Could you add the `type(model)` into the error message? For example,
`'Expected model type is an ONNX ModelProto but got a %s' % type(model)'`

onnxmltools/utils/float16_converter.py
wschin: Agree with @stevenlix . Examples are always not enough (until the examples we have can cover ALL user cases).

onnxmltools/utils/float16_converter.py
```diff
@@ -17,16 +22,28 @@ def convert_float_to_float16(model):
...
     ::
 
+        Example 1: Convert ONNX model ProtoBuf object:
+        from onnxmltools.utils.float16_converter import convert_float_to_float16
+        new_onnx_model = convert_float_to_float16(onnx_model)
+
+        Example 2: Convert ONNX model binary file:
         from onnxmltools.utils.float16_converter import convert_float_to_float16
+        from onnxmltools.utils import load_model, save_model
+        onnx_model = load_model("model.onnx")
```
wschin: Many places we use ' instead of "

onnxmltools/utils/float16_converter.py
```diff
@@ -17,16 +22,28 @@ def convert_float_to_float16(model):
     '''
     Convert TensorProto DataType float in the ONNX model to float16.
 
-    :param model: ONNX model
-    :return: converted ONNX model
+    :param model: ONNX model ProtoBuf object
+    :return: converted ONNX model ProtoBuf object
 
     Example:
```
wschin: Example"s".

PR #99, opened 2018-06-26
Update coreml converter classifier and regressor output shape.

wenbingl:
For the tree ensemble classifier (and this is true for other classifiers as well), one of the outputs has shape {N}, but the converter uses {N,1}. This fails some ONNX engine shape verification.

### DIFF
--- a/onnxmltools/convert/coreml/shape_calculators/Classifier.py
+++ b/onnxmltools/convert/coreml/shape_calculators/Classifier.py
@@ -43,7 +43,7 @@ def calculate_traditional_classifier_output_shapes(operator):
     if operator.targeted_onnx_version < StrictVersion('1.2'):
         output_shape = [1, 1]
     else:
-        output_shape = [N, 1]
+        output_shape = [N]
 
     if class_label_type == 'stringClassLabels':
         operator.outputs[0].type = StringTensorType(output_shape, doc_string=operator.outputs[0].type.doc_string)
--- a/onnxmltools/convert/coreml/shape_calculators/Classifier.py
+++ b/onnxmltools/convert/coreml/shape_calculators/Classifier.py
@@ -14,8 +14,8 @@ from ...common.utils import check_input_and_output_numbers, check_input_and_outp
 def calculate_traditional_classifier_output_shapes(operator):
     '''
     For classifiers, allowed input/output patterns are
-        1. [N, C_1], ..., [N, C_n] ---> [N, 1], Sequence of Map
-        2. [N, C_1], ..., [N, C_n] ---> [N, 1]
+        1. [N, C_1], ..., [N, C_n] ---> [N], Sequence of Map
+        2. [N, C_1], ..., [N, C_n] ---> [N]
 
     For regressors, allowed input/output patterns are
         1. [N, C_1], ..., [N, C_n] ---> [N, 1]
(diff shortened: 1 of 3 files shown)


wenbingl:
For the tree ensemble classifier (and this is true for other classifiers as well), one of the outputs has shape {N}, but the converter uses {N,1}. This fails some ONNX engine shape verification.

onnxmltools/convert/coreml/shape_calculators/Regressor.py
```diff
@@ -34,8 +35,10 @@ def calculate_traditional_regressor_output_shapes(operator):
         raise ValueError('Model should be one of linear model, tree-based model, and support vector machine')
 
     N = operator.inputs[0].type.shape[0]
-    operator.outputs[0].type = FloatTensorType([N, C], doc_string=operator.outputs[0].type.doc_string)
-
+    if operator.targeted_onnx_version < StrictVersion('1.2') or C > 1:
+        operator.outputs[0].type = FloatTensorType([N, C], doc_string=operator.outputs[0].type.doc_string)
+    else:
+        operator.outputs[0].type = FloatTensorType([N], doc_string=operator.outputs[0].type.doc_string)
```
wschin: There are some more files which may need to be changed. For example,  scikit-learn [linear classifier](https://github.com/onnx/onnxmltools/blob/master/onnxmltools/convert/sklearn/shape_calculators/LinearClassifier.py), [SVM](https://github.com/onnx/onnxmltools/blob/master/onnxmltools/convert/sklearn/shape_calculators/SVM.py).

onnxmltools/convert/coreml/shape_calculators/Classifier.py
```diff
@@ -43,7 +43,7 @@ def calculate_traditional_classifier_output_shapes(operator):
     if operator.targeted_onnx_version < StrictVersion('1.2'):
         output_shape = [1, 1]
     else:
-        output_shape = [N, 1]
+        output_shape = [N]
```
wschin: I believe this new bahavior needs to be clarified in ONNX spec as well. Basically, the current spec says nothing about this. If it doesn't match previous ONNX behavior, version number should be increased too.

onnxmltools/convert/coreml/shape_calculators/Regressor.py
wenbingl: ok, but I am a little doubting that whether onnx can support [N, C] output where C > 1.

Outputs
Y : tensor(float)
N classes

wschin: It's not documented, so anything can happen. Maybe double-check with @linkerzhang and clarify in ONNX spec. Please make sure the final definition you reach doesn't not violate the released runtime if possible.

onnxmltools/convert/coreml/shape_calculators/Classifier.py
wschin: One reason to have 2-D is to make sure regressor and classifier have the same shape interface. For multi-output regression, its output shape must be [N, C]. If we consider classifier is a special case of regression, its output shape may be [N, 1] here.

wschin:
Did you run some tests with previous runtime by any chance?

wenbingl:
We have a work item to run all model test on onnx == 1.0 for RS4 to make sure target_onnx_version argument really works on both runtimes.

PR #106, opened 2018-07-08
LightGBM/ONNX conversion improvements

gpminsuk:
- Added FeatureVectorizer operation when input is more than 1
- ZipMap operation only for multi-class LGBMClassifier
- Fixed label output type bug ([N, 1] -> [N])
- Fixed probabilities output type bug ([N, 1] -> [N, E])

### DIFF
--- a/onnxmltools/convert/sklearn/operator_converters/LightGbm.py
+++ b/onnxmltools/convert/sklearn/operator_converters/LightGbm.py
@@ -179,11 +179,11 @@ def convert_lightgbm(scope, operator, container):
             raise ValueError('Only string and integer class labels are allowed')
 
         if len(class_labels) > 2 and operator.type != 'SklearnLinearSVC':
-            probability_tensor_name = scope.get_unique_variable_name('probability_tensor')
             # Create tree classifier
+            probability_tensor_name = scope.get_unique_variable_name('probability_tensor')
             container.add_node('TreeEnsembleClassifier', feature_name,
-                            [operator.outputs[0].full_name, probability_tensor_name],
-                            op_domain='ai.onnx.ml', **attrs)
+                               [operator.outputs[0].full_name, probability_tensor_name],
+                               op_domain='ai.onnx.ml', **attrs)
 
             # Convert probability tensor to probability map (keys are labels while values are the associated probabilities)
             container.add_node('ZipMap', probability_tensor_name, operator.outputs[1].full_name,
@@ -191,7 +191,7 @@ def convert_lightgbm(scope, operator, container):
         else:
             # Create tree classifier
             container.add_node('TreeEnsembleClassifier', feature_name,
-                           operator.output_full_names, op_domain='ai.onnx.ml', **attrs)
+                               operator.output_full_names, op_domain='ai.onnx.ml', **attrs)
 
     else:
         # Create tree regressor
--- a/onnxmltools/convert/sklearn/operator_converters/LightGbm.py
+++ b/onnxmltools/convert/sklearn/operator_converters/LightGbm.py
@@ -10,6 +10,7 @@ import numpy as np
 from collections import Counter
 from lightgbm import LGBMClassifier, LGBMRegressor
 from ...common._registration import register_converter
+from .common import concatenate_variables
 from .TreeEnsemble import _get_default_tree_classifier_attribute_pairs
 
 
@@ -117,6 +118,11 @@ def convert_lightgbm(scope, operator, container):
     if gbm_model.boosting_type != 'gbdt':
         raise ValueError('Only support LightGBM classifier with boosting_type=gbdt, current boosting_type=' + gbm_model.boosting_type)
     gbm_text = gbm_model.booster_.dump_model()
+    
+    if len(operator.inputs) > 1:
+        feature_name = concatenate_variables(scope, operator.inputs, container)
+    else:
+        feature_name = operator.inputs[0].full_name
 
     attrs = _get_default_tree_classifier_attribute_pairs()
     attrs['name'] = operator.full_name
@@ -173,15 +179,15 @@ def convert_lightgbm(scope, operator, container):
             raise ValueError('Only string and integer class labels are allowed')
 
         # Create tree classifier
-        label_tensor_name = scope.get_onnx_variable_name('label')
-        probability_tensor_name = scope.get_onnx_variable_name('Probability')
-        container.add_node('TreeEnsembleClassifier', operator.input_full_names,
-                           [label_tensor_name, probability_tensor_name],
+        
+        probability_tensor_name = scope.get_unique_variable_name('probability_tensor')
+        container.add_node('TreeEnsembleClassifier', feature_name,
+                           [operator.outputs[0].full_name, probability_tensor_name],
                            op_domain='ai.onnx.ml', **attrs)
 
         # Convert probability tensor to probability map (keys are labels while values are the associated probabilities)
-        #container.add_node('ZipMap', probability_tensor_name, operator.outputs[1].full_name,
-        #                   op_domain='ai.onnx.ml', **zipmap_attrs)
+        container.add_node('ZipMap', probability_tensor_name, operator.outputs[1].full_name,
+                           op_domain='ai.onnx.ml', **zipmap_attrs)
     else:
         # Create tree regressor
         keys_to_be_renamed = list(k for k in attrs.keys() if k.startswith('class_'))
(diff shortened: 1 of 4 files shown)


onnxmltools/convert/main.py
```diff
@@ -8,12 +8,12 @@
 from .common import utils
 
 
-def convert_sklearn(model, name=None, initial_types=None, doc_string='', targeted_onnx=onnx.__version__):
+def convert_sklearn(model, name=None, initial_types=None, doc_string='', targeted_onnx=onnx.__version__, label_variable_name=None, probabilities_variable_name=None):
```
wenbingl: any user case or particular reason to make this API change?

onnxmltools/convert/sklearn/shape_calculators/LinearClassifier.py
```diff
@@ -19,43 +19,33 @@ def calculate_sklearn_linear_classifier_output_shapes(operator):
...
     N = operator.inputs[0].type.shape[0]
 
     class_labels = operator.raw_operator.classes_
     if all(isinstance(i, np.ndarray) for i in class_labels):
         class_labels = np.concatenate(class_labels)
     if all(isinstance(i, (six.string_types, six.text_type)) for i in class_labels):
         operator.outputs[0].type = StringTensorType(shape=[N, 1])
-        if len(class_labels) > 2 or operator.type != 'SklearnLinearSVC':
-            # For multi-class classifier, we produce a map for encoding the probabilities of all classes
-            if operator.targeted_onnx_version < StrictVersion('1.2'):
```
wenbingl: why is the backward compatibility code deleted?

onnxmltools/convert/main.py
gpminsuk: PredictionUnit API in Windows 10 requires specific output name "Probability" instead of "probabilities". I am open to nicer design changes but this was easiest fix.

onnxmltools/convert/sklearn/shape_calculators/LinearClassifier.py
gpminsuk: I was asked to convert LightGBM multi-class model also to ONNX for PredictionUnit API in Windows 10. I am not sure why this was written to output SequenceType. Was there any particular reason?

onnxmltools/convert/main.py
wenbingl: This is a special requirement, which can't be adopted for a general API design.

Why not directly modify the output name after the model converted.

onnxmltools/convert/sklearn/shape_calculators/LinearClassifier.py
wenbingl: it complys with the onnx spec https://github.com/onnx/onnx/blob/master/docs/Operators-ml.md#ai.onnx.ml.ZipMap

onnxmltools/convert/main.py
gpminsuk: I am not adding anything about my special requirements. What I added is capability to configure hard-coded output names. However, it is good a suggestion to do it outside. How can I rename output node after conversion?

onnxmltools/convert/sklearn/shape_calculators/LinearClassifier.py
gpminsuk: Yes, I know ZipMap complies onnx spec. I am asking why do you do ZipMap operation after TreeEnsembleClassifier?

wenbingl:
Can you connect to Microsoft ONNX team or PM for the more details?

onnxmltools/convert/main.py
wschin: It's a bit strange to specify label_variable_name or probabilities_variable_name. There is no formal definitions for them! Let's think about an ensemble model, a pipeline with one linear classifier and random forest classifier. Which variable should be associated with label_variable_name/probabilities_variable_name?

A basic idea for modifying ONNX model in python (assume that your model variable is `model`: 
  1. Iterate through `model.graph.node` and change all the names of being targeted. 
  2. Also, modify names in `model.output` accordingly

onnxmltools/convert/sklearn/shape_calculators/LinearClassifier.py
wschin: It's an idea from Core ML. I personally think it's nice to associate label names and their probabilities from user perspective. On the other hand, it's doable to remove all ZipMap's in scikit-learn conversions but test data should be modified as well.

gpminsuk: I reached out to Microsoft ONNX team and they said they use ZipMap to be consistent on Tree implementation in scikit-learn (i.e., all tree models return probabilities as map not [N, E]). Following up with them more about how I used this tool for our requirements.

onnxmltools/convert/sklearn/operator_converters/LightGbm.py
```diff
@@ -115,8 +116,13 @@ def _parse_node(tree_id, class_id, node_id, node_id_pool, learning_rate, node, a
 def convert_lightgbm(scope, operator, container):
     gbm_model = operator.raw_operator
     if gbm_model.boosting_type != 'gbdt':
-        raise ValueError('Only support LightGBM classifier with boosting_type=gbdt')
+        raise ValueError('Only support LightGBM classifier with boosting_type=gbdt, requested boosting_type=' + gbm_model.boosting_type)
```
wschin: `'Only support LightGBM classifier with boosting_type=gbdt, requested boosting_type=' + gbm_model.boosting_type` --> `'Only support LightGBM classifier with boosting_type=gbdt, but got boosting_type=' + gbm_model.boosting_type'`

onnxmltools/convert/sklearn/operator_converters/LightGbm.py
```diff
@@ -115,8 +116,13 @@ def _parse_node(tree_id, class_id, node_id, node_id_pool, learning_rate, node, a
...
     gbm_text = gbm_model.booster_.dump_model()
+    
+    if len(operator.inputs) > 1:
```
wschin: Does LGBM scikit-learn interface support multiple input feature vectors?

onnxmltools/convert/sklearn/operator_converters/LinearClassifier.py
```diff
@@ -66,7 +66,7 @@ def convert_sklearn_linear_classifier(scope, operator, container):
         normalized_probability_tensor_name = probability_tensor_name
 
     # Post-process probability tensor produced by LinearClassifier operator
-    if len(class_labels) > 2 or op.__class__.__name__ != 'LinearSVC':
+    if len(class_labels) > 2 and op.__class__.__name__ != 'LinearSVC':
```
wschin: Why do we need to use `and`? `LinearSVC` has no probability output no matter how many classes do we have. Therefore, probability normalization in the else-branch below should be avoided. If this change is necessary, I'd suggest to open another PR to isolate unrelated fixes.

onnxmltools/convert/sklearn/shape_calculators/LinearClassifier.py
```diff
@@ -19,43 +19,33 @@ def calculate_sklearn_linear_classifier_output_shapes(operator):
...
-        1. [N, C] ---> [N, 1], A sequence of map
+        1. [N, C] ---> [N], [N, E] (binary-class)
+        2. [N, 1], ..., [N, 1] ---> [N], [N, E] (binary-class)
+        3. [N, C] ---> [N], A sequence of map (multi-class)
+        4. [N, 1], ..., [N, 1] ---> [N], A sequence of map (multi-class)
 
     Note that the second case is not allowed as long as ZipMap only produces dictionary.
     '''
-    check_input_and_output_numbers(operator, input_count_range=1, output_count_range=[1, 2])
+    check_input_and_output_numbers(operator, input_count_range=[1, None], output_count_range=[1, 2])
```
wschin: I don't think scikit-learn can support multiple inputs. Is it required by other packages?

onnxmltools/convert/sklearn/shape_calculators/LinearClassifier.py
```diff
@@ -19,43 +19,33 @@ def calculate_sklearn_linear_classifier_output_shapes(operator):
...
-        if len(class_labels) > 2 or operator.type != 'SklearnLinearSVC':
-            # For multi-class classifier, we produce a map for encoding the probabilities of all classes
-            if operator.targeted_onnx_version < StrictVersion('1.2'):
-                operator.outputs[1].type = DictionaryType(Int64TensorType([1]), FloatTensorType([1]))
-            else:
-                operator.outputs[1].type = SequenceType(DictionaryType(Int64TensorType([]), FloatTensorType([])), N)
-        else:
-            # For binary classifier, we produce the probability of the positive class
-            operator.outputs[1].type = FloatTensorType(shape=[N, 1])
+        operator.outputs[1].type = FloatTensorType(shape=[N, len(class_labels)])
```
wschin: Looks like dropping ZipMap can make the code cleaner. On the other hand, this change may incur changes on test side, so we need to sync with test owners to see if they are ok.

PR #110, opened 2018-07-17
Fix the keras reshape op conversion.

wenbingl:
according to keras spec,  target_shape does not include the batch axis.


onnxmltools/convert/keras/operator_converters/Reshape.py
```diff
@@ -12,7 +12,7 @@
 def convert_keras_reshape(scope, operator, container):
     op = operator.raw_operator
     apply_reshape(scope, operator.inputs[0].full_name, operator.outputs[0].full_name, container,
-                  desired_shape=op.target_shape)
+                  desired_shape=[-1, ] + list(op.target_shape))
```
wschin: `[-1, ]` --> `[-1]`

PR #114, opened 2018-07-21
The custom operator support for the keras/coreml/sklearn converters. 

wenbingl:


wenbingl:
The previous PR was polluted by another commit, will drop it.

PR #115, opened 2018-07-24
Manipulation of metadata props

tiagoshibata:
* Add functions to manipulate metadata props
* Add metadata props automatically if the input model has an image input

### DIFF
--- a/onnxmltools/convert/common/_topology.py
+++ b/onnxmltools/convert/common/_topology.py
@@ -6,7 +6,7 @@
 
 import re
 from distutils.version import StrictVersion
-from ...utils.metadata_props import add_metadata_props, set_denotation
+from ...utils.metadata_props import add_metadata_props, color_space_to_pixel_format, set_denotation
 from ...proto import onnx
 from ...proto import helper
 from .data_types import *
@@ -664,12 +664,7 @@ def convert_topology(topology, model_name, doc_string, targeted_onnx):
         color_space = getattr(variable.type, 'color_space', None)
         if color_space:
             image_inputs.append(variable.raw_name)
-            color_space_to_pixel_format = {
-                'BGR': 'Bgr8',
-                'RGB': 'Rgb8',
-                'GRAY': 'Gray8',
-            }
-            pixel_format = color_space_to_pixel_format[color_space]
+            pixel_format = color_space_to_pixel_format(color_space)
             if metadata_props.get('Image.BitmapPixelFormat', pixel_format) != pixel_format:
                 print('Warning: conflicting pixel formats found. In ONNX, all input/output images must use the same pixel format.')
             metadata_props = {
@@ -744,7 +739,8 @@ def convert_topology(topology, model_name, doc_string, targeted_onnx):
     # Add extra information
     add_metadata_props(onnx_model, metadata_props)
     for image in image_inputs:
-        set_denotation(onnx_model, image, 'IMAGE', dimension_denotation=['DATA_BATCH', 'DATA_CHANNEL', 'DATA_FEATURE', 'DATA_FEATURE'])
+        set_denotation(onnx_model, image, 'IMAGE',
+                       dimension_denotation=['DATA_BATCH', 'DATA_CHANNEL', 'DATA_FEATURE', 'DATA_FEATURE'])
     onnx_model.ir_version = onnx_proto.IR_VERSION
     onnx_model.producer_name = utils.get_producer()
     onnx_model.producer_version = utils.get_producer_version()
--- a/onnxmltools/convert/common/_topology.py
+++ b/onnxmltools/convert/common/_topology.py
@@ -8,7 +8,7 @@ import re
 from distutils.version import StrictVersion
 from ...proto import onnx
 from ...proto import helper
-from ...utils.metadata_props import add_metadata_props, color_space_to_pixel_format, set_denotation
+from ...utils.metadata_props import add_metadata_props
 from .data_types import *
 from ._container import ModelComponentContainer
 from . import _registration
@@ -234,7 +234,7 @@ class Topology:
 
     def __init__(self, model, default_batch_size=1, initial_types=None,
                  reserved_variable_names=None, reserved_operator_names=None, targeted_onnx=None,
-                 custom_conversion_functions=None, custom_shape_calculators=None):
+                 custom_conversion_functions=None, custom_shape_calculators=None, metadata_props={}):
         '''
         Initialize a Topology object, which is an intermediate representation of a computational graph.
 
@@ -257,6 +257,7 @@ class Topology:
         self.targeted_onnx_version = StrictVersion(targeted_onnx)
         self.custom_conversion_functions = custom_conversion_functions if custom_conversion_functions else {}
         self.custom_shape_calculators = custom_shape_calculators if custom_shape_calculators else {}
+        self.metadata_props = metadata_props
 
         # This attribute is used in optimizing the graph structure. If root_names is not empty, only the variables
         # specified will be treated as the roots (i.e., set is_fed to True in the beginning of a graph evaluation) of
@@ -657,22 +658,6 @@ def convert_topology(topology, model_name, doc_string, targeted_onnx):
         if name in other_inputs:
             container.add_input(other_inputs[name])
 
-    # Add metadata properties and channel denotation to image tensors
-    metadata_props = {}
-    image_inputs = []
-    for variable in topology.find_root_and_sink_variables():
-        color_space = getattr(variable.type, 'color_space', None)
-        if color_space:
-            image_inputs.append(variable.raw_name)
-            pixel_format = color_space_to_pixel_format(color_space)
-            if metadata_props.get('Image.BitmapPixelFormat', pixel_format) != pixel_format:
-                print('Warning: conflicting pixel formats found. In ONNX, all input/output images must use the same pixel format.')
-            metadata_props = {
-                'Image.BitmapPixelFormat': pixel_format,
-                'Image.ColorSpaceGamma': 'SRGB',
-                'Image.NominalPixelRange': 'NominalRange_0_255',
-            }
-
     # Add leaves the graph according to their order in the original model
     for name in topology.raw_model.output_names:
         if name in tensor_outputs:
@@ -737,10 +722,7 @@ def convert_topology(topology, model_name, doc_string, targeted_onnx):
         i += 1
 
     # Add extra information
-    add_metadata_props(onnx_model, metadata_props)
-    for image in image_inputs:
-        set_denotation(onnx_model, image, 'IMAGE',
-                       dimension_denotation=['DATA_BATCH', 'DATA_CHANNEL', 'DATA_FEATURE', 'DATA_FEATURE'])
+    add_metadata_props(onnx_model, topology.metadata_props)
     onnx_model.ir_version = onnx_proto.IR_VERSION
     onnx_model.producer_name = utils.get_producer()
     onnx_model.producer_version = utils.get_producer_version()
--- a/onnxmltools/convert/common/_topology.py
+++ b/onnxmltools/convert/common/_topology.py
@@ -6,9 +6,9 @@
 
 import re
 from distutils.version import StrictVersion
-from ...utils.metadata_props import add_metadata_props, color_space_to_pixel_format, set_denotation
 from ...proto import onnx
 from ...proto import helper
+from ...utils.metadata_props import add_metadata_props, color_space_to_pixel_format, set_denotation
 from .data_types import *
 from ._container import ModelComponentContainer
 from . import _registration
--- a/onnxmltools/convert/common/_topology.py
+++ b/onnxmltools/convert/common/_topology.py
@@ -6,6 +6,7 @@
 
 import re
 from distutils.version import StrictVersion
+from ...utils.metadata_props import add_metadata_props
 from ...proto import onnx
 from ...proto import helper
 from .data_types import *
@@ -651,15 +652,24 @@ def convert_topology(topology, model_name, doc_string, targeted_onnx):
 
     # Add roots the graph according to their order in the original model
     for name in topology.raw_model.input_names:
-        tensor = tensor_inputs.get(name)
-        if tensor:
-            container.add_input(tensor)
-            if tensor.type.denotation == 'IMAGE':
-                metadata_props.update({
-                    'Image.BitmapPixelFormat': 'Bgr8',
-                    'Image.ColorSpaceGamma': 'SRGB',
-                    'Image.NominalPixelRange': 'NominalRange_0_255',
-                })
+        if name in tensor_inputs:
+            container.add_input(tensor_inputs[name])
+    for variable in topology.find_root_and_sink_variables():
+        color_space = getattr(variable.type, 'color_space', None)
+        if color_space:
+            color_space_to_pixel_format = {
+                'BGR': 'Bgr8',
+                'RGB': 'Rgb8',
+                'GRAY': 'Gray8',
+            }
+            pixel_format = color_space_to_pixel_format[color_space]
+            if metadata_props.get('Image.BitmapPixelFormat', pixel_format) != pixel_format:
+                print('Warning: conflicting pixel formats found. In ONNX, all input/output images must use the same pixel format.')
+            metadata_props = {
+                'Image.BitmapPixelFormat': pixel_format,
+                'Image.ColorSpaceGamma': 'SRGB',
+                'Image.NominalPixelRange': 'NominalRange_0_255',
+            }
     for name in topology.raw_model.input_names:
         if name in other_inputs:
             container.add_input(other_inputs[name])
@@ -728,8 +738,7 @@ def convert_topology(topology, model_name, doc_string, targeted_onnx):
         i += 1
 
     # Add extra information
-    onnx_model.metadata_props.extend(onnx_proto.StringStringEntryProto(key=key, value=value)
-                                     for key, value in metadata_props.items())
+    add_metadata_props(onnx_model, metadata_props)
     onnx_model.ir_version = onnx_proto.IR_VERSION
     onnx_model.producer_name = utils.get_producer()
     onnx_model.producer_version = utils.get_producer_version()
--- a/onnxmltools/convert/common/_topology.py
+++ b/onnxmltools/convert/common/_topology.py
@@ -629,6 +629,7 @@ def convert_topology(topology, model_name, doc_string, targeted_onnx):
     topology._initialize_graph_status_for_traversing()
 
     container = ModelComponentContainer(targeted_onnx)
+    metadata_props = {}
 
     # Put roots and leaves as ONNX's model into buffers. They will be added into ModelComponentContainer later.
     tensor_inputs = {}
@@ -650,8 +651,15 @@ def convert_topology(topology, model_name, doc_string, targeted_onnx):
 
     # Add roots the graph according to their order in the original model
     for name in topology.raw_model.input_names:
-        if name in tensor_inputs:
-            container.add_input(tensor_inputs[name])
+        tensor = tensor_inputs.get(name)
+        if tensor:
+            container.add_input(tensor)
+            if tensor.type.denotation == 'IMAGE':
+                metadata_props.update({
+                    'Image.BitmapPixelFormat': 'Bgr8',
+                    'Image.ColorSpaceGamma': 'SRGB',
+                    'Image.NominalPixelRange': 'NominalRange_0_255',
+                })
     for name in topology.raw_model.input_names:
         if name in other_inputs:
             container.add_input(other_inputs[name])
@@ -720,6 +728,8 @@ def convert_topology(topology, model_name, doc_string, targeted_onnx):
         i += 1
 
     # Add extra information
+    onnx_model.metadata_props.extend(onnx_proto.StringStringEntryProto(key=key, value=value)
+                                     for key, value in metadata_props.items())
     onnx_model.ir_version = onnx_proto.IR_VERSION
     onnx_model.producer_name = utils.get_producer()
     onnx_model.producer_version = utils.get_producer_version()
--- a/onnxmltools/convert/common/_topology.py
+++ b/onnxmltools/convert/common/_topology.py
@@ -234,7 +234,7 @@ class Topology:
 
     def __init__(self, model, default_batch_size=1, initial_types=None,
                  reserved_variable_names=None, reserved_operator_names=None, targeted_onnx=None,
-                 custom_conversion_functions=None, custom_shape_calculators=None, metadata_props={}):
+                 custom_conversion_functions=None, custom_shape_calculators=None, metadata_props=None):
         '''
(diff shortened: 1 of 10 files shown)


onnxmltools/convert/common/_topology.py
```diff
@@ -639,6 +640,22 @@ def convert_topology(topology, model_name, doc_string, targeted_onnx):
...
+    metadata_props = {}
+    image_inputs = []
+    for variable in topology.find_root_and_sink_variables():
+        color_space = getattr(variable.type, 'color_space', None)
+        if color_space:
+            image_inputs.append(variable.raw_name)
+            pixel_format = color_space_to_pixel_format(color_space)
+            if metadata_props.get('Image.BitmapPixelFormat', pixel_format) != pixel_format:
+                print('Warning: conflicting pixel formats found. In ONNX, all input/output images must use the same pixel format.')
+            metadata_props = {
```
wenbingl: these default value aligns with Core ML, or other standard?

onnxmltools/convert/common/utils.py
```diff
@@ -213,6 +213,12 @@ def convert_to_list(var):
         raise TypeError('Unable to flatten variable')
 
 
+def version_string_to_tuple(version):
+    version = version.split('.')
```
wenbingl: using StrictVersion to keep consistent with other functions.

onnxmltools/utils/metadata_props.py
```diff
@@ -0,0 +1,67 @@
...
+    'GRAY': 'Gray8',
+}
+
+
+def color_space_to_pixel_format(color_space):
+    return COLOR_SPACE_TO_PIXEL_FORMAT[color_space]
+
+
+def add_metadata_props(onnx_model, metadata_props, targeted_onnx=onnx.__version__):
+    if version_string_to_tuple(targeted_onnx) < (1, 2, 1) or onnx_model.ir_version < 3:
```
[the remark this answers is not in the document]
wenbingl: this should be fine.

wenbingl: Are sure the value of onnx_model.ir_version was already assigned on calling add_metadata_props.

onnxmltools/convert/common/_topology.py
tiagoshibata: SRGB and 0-255 are the most commonly used formats for input images. They are also the values used in type denotation sample in the ONNX documentation. We believe it's a sane default if the user doesn't provide any information.

tiagoshibata:
Ping, let us know what you think about our latest changes.

wenbingl:
please add an unit test for that.

onnxmltools/convert/common/_topology.py
```diff
@@ -6,6 +6,7 @@
 
 import re
 from distutils.version import StrictVersion
+from ...utils.metadata_props import add_metadata_props, color_space_to_pixel_format, set_denotation
```
wschin: Please move `from ...utils.metadata_props import add_metadata_props, color_space_to_pixel_format, set_denotation` to line 12 (right after `from ...proto import helper`).

onnxmltools/utils/metadata_props.py
```diff
@@ -0,0 +1,68 @@
...
+    for prop in model_metadata:
+        if prop.key.lower() in new_props:
+            model_metadata.remove(prop)
+    model_metadata.extend(onnx_proto.StringStringEntryProto(key=key, value=value)
+                                     for key, value in metadata_props.items())
+
+
+def set_denotation(onnx_model, input_name, denotation, dimension_denotation=None, targeted_onnx=onnx.__version__):
+    if StrictVersion(targeted_onnx) < StrictVersion('1.2.1'):
+        print('Metadata properties are not supported in this model')
```
wschin: '...in this model' ---> '...in the targeted ONNX-%s' % targeted_onnx

onnxmltools/convert/common/_topology.py
wschin: Are those information (e.g., `Image.BitmapPixelFormat`) formal definitions in ONNX? If no, I'd prefer to leave this job (adding metadata strings) to the parser for source models (e.g., for Core ML models, we have [this](https://github.com/onnx/onnxmltools/blob/f8743958da9594488c29fdffb501f53f2f5ecc5d/onnxmltools/convert/coreml/_parse.py#L49))

wschin: Also, this metadata dictionary gets overwritten if there are multiple input images..

onnxmltools/convert/common/_topology.py
```diff
@@ -700,6 +717,10 @@ def convert_topology(topology, model_name, doc_string, targeted_onnx):
         i += 1
 
     # Add extra information
+    add_metadata_props(onnx_model, metadata_props)
+    for image in image_inputs:
+        set_denotation(onnx_model, image, 'IMAGE',
```
wschin: This may be true to Core ML. If things comes from Keras, it's not necessary because Keras can have `channels_last=True` where `DATA_CHANNEL` should be moved to the tail of the denotation list. To resolve this problem, we need to add a denotation list into our `[TensorType](https://github.com/onnx/onnxmltools/blob/f8743958da9594488c29fdffb501f53f2f5ecc5d/onnxmltools/convert/common/data_types.py#L56)` and maintain it via the parser.

onnxmltools/utils/metadata_props.py
```diff
@@ -0,0 +1,68 @@
+from distutils.version import StrictVersion
+
```
wschin: Please remove the empty line here and move `from distutils.version import StrictVersion ` to line 4.

onnxmltools/convert/common/_topology.py
wschin: I see. Thanks. Please see also my comment for `add_metadata_props(onnx_model, metadata_props)` below. My feeling is that we should add such information at the earliest place, which is the parsing function for different frameworks (e.g., Core ML and scikit-learn) and maintain it throughout the compiling procedure. Although I am still open to your current approach because Core ML is the only model which needs medata.

onnxmltools/convert/common/_topology.py
tiagoshibata: Yes, something like you are describing was in our initial proposal ([see the first commit in this PR](https://github.com/onnx/onnxmltools/pull/115/commits/e5bff90ec457078041f7b0cc9b4edfeb8bcd1650#diff-8ddffe74cd10c33d49c52adee4f9e22b)), but it was suggested that we don't clutter `TensorType` with fields that are currently CoreML specific. I believe our previous approach (the one you are suggesting) is better, should I go back to that?

tiagoshibata: I'll add denotation and channel_denotations parameters to TensorType

wschin: You will probably need to propagate such information throughout the shape inference stage. This is necessary for some edge cases (our compiling procedure removes identity by replacing its output with its input. If we don't copy metadata from the input to the output in shape inference stage, we will lose it).

onnxmltools/utils/metadata_props.py
```diff
@@ -0,0 +1,67 @@
+from ..proto import onnx, onnx_proto
+from distutils.version import StrictVersion
+
+
+def _get_case_insensitive(iterable, key):
+    return next((x for x in iterable if x.lower() == key.lower()), None)
+
+
+def _validate_metadata(metadata_props):
```
wschin: Please add `# metadata_props is a Python dictionary. For each key-value pair, the key and value specify a variable's name and its actual value, respectively.` This might help potential devs to know your work more.

onnxmltools/utils/metadata_props.py
```diff
@@ -0,0 +1,67 @@
...
+    valid_metadata_props = {
+        'Image.BitmapPixelFormat': ['Gray8', 'Rgb8', 'Bgr8', 'Rgba8', 'Bgra8'],
+        'Image.ColorSpaceGamma': ['Linear', 'SRGB'],
+        'Image.NominalPixelRange': ['NominalRange_0_255', 'Normalized_0_1', 'Normalized_1_1', 'NominalRange_16_235'],
+    }
+    image_metadata_props = {k: v for k, v in metadata_props.items() if _get_case_insensitive(valid_metadata_props, k)}
+    for key, value in image_metadata_props.items():
+        key = _get_case_insensitive(valid_metadata_props, key)
+        if not key:
+            print('Warning: key {} is defined multiple times'.format(key))
```
wschin: Better to use `warning` module in python to create warning message. I am also not sure if it's just a warning or an exception.

onnxmltools/utils/metadata_props.py
```diff
@@ -0,0 +1,67 @@
...
+        else:
+            valid_values = valid_metadata_props.pop(key)
+            if not _get_case_insensitive(valid_values, value):
+                print('Warning: value {} is invalid. Valid values are {}'.format(value, valid_values))
```
wschin: Better to use warning module in python to create warning message.

onnxmltools/utils/metadata_props.py
```diff
@@ -0,0 +1,67 @@
...
+    if image_metadata_props and valid_metadata_props:
+        print('Warning: incomplete image metadata is being added. Keys {} are missing.'.format(', '.join(valid_metadata_props)))
+
```
wschin: Only one line between a function and the following dictionary?

onnxmltools/utils/metadata_props.py
```diff
@@ -0,0 +1,67 @@
...
+COLOR_SPACE_TO_PIXEL_FORMAT = {
```
wschin: We can use `Bgr8`, `Rgb8`, and `Gray8` to replace those ones currently used. Then, we can get rid of the following translation.

onnxmltools/utils/metadata_props.py
```diff
@@ -0,0 +1,67 @@
...
+}
+
+
+def color_space_to_pixel_format(color_space):
+    return COLOR_SPACE_TO_PIXEL_FORMAT[color_space]
+
+
+def add_metadata_props(onnx_model, metadata_props, targeted_onnx=onnx.__version__):
+    if StrictVersion(targeted_onnx) < StrictVersion('1.2.1'):
+        print('Metadata properties are not supported in targeted ONNX-%s' % targeted_onnx)
```
wschin: It's an exception. Please throw one.

onnxmltools/utils/metadata_props.py
```diff
@@ -0,0 +1,67 @@
...
+        return
+    _validate_metadata(metadata_props)
+    # Overwrite old properties (case insensitive)
+    new_props = [x.lower() for x in metadata_props]
+    model_metadata = onnx_model.metadata_props
+    for prop in model_metadata:
+        if prop.key.lower() in new_props:
+            model_metadata.remove(prop)
```
wschin: Why there is a possibility that we add something and then remove it? Could this be avoided?

onnxmltools/utils/metadata_props.py
```diff
@@ -0,0 +1,67 @@
...
+    model_metadata.extend(onnx_proto.StringStringEntryProto(key=key, value=value)
+                                     for key, value in metadata_props.items())
+
+
+def set_denotation(onnx_model, input_name, denotation, dimension_denotation=None, targeted_onnx=onnx.__version__):
```
wschin: Where is it being used? I think `TensorType` already implements essentially the same function.

onnxmltools/utils/metadata_props.py
```diff
@@ -0,0 +1,67 @@
...
+    return next((x for x in iterable if x.lower() == key.lower()), None)
+
+
+def _validate_metadata(metadata_props):
+    valid_metadata_props = {
+        'Image.BitmapPixelFormat': ['Gray8', 'Rgb8', 'Bgr8', 'Rgba8', 'Bgra8'],
+        'Image.ColorSpaceGamma': ['Linear', 'SRGB'],
+        'Image.NominalPixelRange': ['NominalRange_0_255', 'Normalized_0_1', 'Normalized_1_1', 'NominalRange_16_235'],
+    }
+    image_metadata_props = {k: v for k, v in metadata_props.items() if _get_case_insensitive(valid_metadata_props, k)}
```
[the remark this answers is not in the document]
tiagoshibata: It's in the spec. Personally I dislike case insensitive interfaces and wished the spec was different, but... such is life.

Will add comments. If you think of a better/cleaner way of doing the case-insensitive code let me know.

onnxmltools/utils/metadata_props.py
tiagoshibata: This is not used in the converter path. This function is to be used when the user has ONNX models and want to add denotation to the existing model, e.g:

```py
import onnxmltools
from onnxmltools.utils import set_denotation
 
onnx_model = onnxmltools.load_model('model.onnx')
onnx_model = set_denotation(onnx_model, 'Input3', 'IMAGE',
                            dimension_denotation=['DATA_BATCH', 'DATA_CHANNEL', 'DATA_FEATURE', 'DATA_FEATURE'])
onnxmltools.utils.save_model(onnx_model, 'model_with_denotation.onnx')
```

onnxmltools/utils/metadata_props.py
```diff
@@ -0,0 +1,67 @@
+from ..proto import onnx, onnx_proto
```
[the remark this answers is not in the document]
tiagoshibata: `add_metadata_props` and `set_denotation` operate on the ONNX model. I'll move the rest into common.

onnxmltools/utils/metadata_props.py
wschin: I am not sure if there is a scenario where user needs to specify such information. The converter should handle this properly.

onnxmltools/utils/metadata_props.py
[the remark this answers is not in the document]
wschin: But we have full control on those keys (e.g., `Image.ColorSpaceGamma`). If we consistently use, say, `Image.ColorSpaceGamma`, do we have other reasons to do case-insensitive here?

onnxmltools/utils/metadata_props.py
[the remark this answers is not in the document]
wschin: Sorry that I am still not fully sure we need `set_denotation`. One of our original design principle is `only support needs users have encountered`.

onnxmltools/utils/metadata_props.py
tiagoshibata: We need a way to add metadata to non-CoreML models. We also want to add metadata to many existing models in the model zoo. What do you suggest?

onnxmltools/utils/metadata_props.py
tiagoshibata: I don't get your point... Metadata props are serialized in the protobuf as a list of `StringStringEntryProto`. If the user calls the function to set some props, any existing entries with the same name have to be removed. We have to remove duplicate entries before extending the list.

onnxmltools/utils/metadata_props.py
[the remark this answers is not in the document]
tiagoshibata: Not all models are born from the converter. If a user creates a ONNX model using some other tool, and the tool uses e.g. `Image.colorspacegama`, loading into onnxmltools and calling `add_metadata_props` should still work.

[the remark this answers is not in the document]
tiagoshibata: I'll refactor the whole thing and have a `CaseInsensitiveDict` class.

onnxmltools/utils/metadata_props.py
wschin: Well, I assume this function is an internal one so that user will never use it.

onnxmltools/utils/metadata_props.py
wschin: Humm.. I didn't realize that you're considering model zoo. Could they be regenerated by converters?

wschin: One idea. Is it possible to push `set_denotation` into the official ONNX repo? They have helper modules.

tiagoshibata: About adding to converters, I don't think any input format has this information build-in. Metadata props are a new thing with ONNX and currently requires the user to manually specify how the model was trained (so if I used RGB instead of BGR I have to explicitly set it in the ONNX model).

onnxmltools/utils/metadata_props.py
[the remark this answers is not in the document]
tiagoshibata: Isn't ONNX -> ONNX a valid conversion? You have code for float16 conversion in the repo, we're only adding metadata editing features.

onnxmltools/utils/metadata_props.py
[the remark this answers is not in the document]
tiagoshibata: Image metadata props are used in WinML to control how tensors with image denotation are handled. Users will change tensor denotations and image metadata in existing models when using the latest WinML APIs. There will be a public need for it (and currently there is an internal need. A few people have used the code in my fork).

tiagoshibata:
2018-08-07-16 49 29-screenshot

^ Extra metadata props removed.

tests/utils/test_utils.py
```diff
@@ -69,3 +71,12 @@ def test_set_docstring_blank(self):
         self.assertRaises(ValueError, set_model_doc_string, onnx_model.doc_string, "sample")
         set_model_doc_string(onnx_model, "", True)
         self.assertEqual(onnx_model.doc_string, "")
+
+    def test_set_denotation(self):
+        this = os.path.dirname(__file__)
+        onnx_file = os.path.join(this, "models", "coreml_OneHotEncoder_BikeSharing.onnx")
+        onnx_model = load_model(onnx_file)
+        set_denotation(onnx_model, "1", "IMAGE", dimension_denotation=["DATA_FEATURE"])
+        if StrictVersion(onnx.__version__) >= StrictVersion('1.2.1'):
```
wenbingl: using  @unittest.skipIf, instead of running an empty test on a low version onnx.

wenbingl:
please get wschin's approval before your check-in.

onnxmltools/utils/metadata_props.py
```diff
@@ -0,0 +1,57 @@
...
+    'Image.ColorSpaceGamma': ['linear', 'srgb'],
+    'Image.NominalPixelRange': ['nominalrange_0_255', 'normalized_0_1', 'normalized_1_1', 'nominalrange_16_235'],
+})
+
+
+def _validate_metadata(metadata_props):
+    '''
+    Validate metadata properties and possibly show warnings or throw exceptions.
+
+    :param metadata_props: A dictionary of metadata properties, with property names and values
```
wschin: What are typical (key, value) pairs? Could you please add two examples into the doc string? Thanks.

onnxmltools/utils/metadata_props.py
```diff
@@ -0,0 +1,57 @@
...
+    new_metadata = CaseInsensitiveDict({x.key: x.value for x in onnx_model.metadata_props})
+    new_metadata.update(metadata_props)
+    del onnx_model.metadata_props[:]
+    onnx_model.metadata_props.extend(
+        onnx_proto.StringStringEntryProto(key=key, value=value)
+        for key, value in metadata_props.items()
+    )
+
+
+def set_denotation(onnx_model, input_name, denotation, dimension_denotation=None, targeted_onnx=onnx.__version__):
```
wschin: Please add a doc string to describe the types of the input arguments and some actual examples.

onnxmltools/utils/metadata_props.py
```diff
@@ -0,0 +1,57 @@
...
+    if len(CaseInsensitiveDict(metadata_props)) != len(metadata_props):
+        raise RuntimeError('Duplicate metadata props found')
+
+    for key, value in metadata_props.items():
+        valid_values = KNOWN_METADATA_PROPS.get(key)
+        if valid_values and value.lower() not in valid_values:
+            warnings.warn('Key {} has invalid value {}. Valid values are {}'.format(key, value, valid_values))
+
+
+def add_metadata_props(onnx_model, metadata_props, targeted_onnx=onnx.__version__):
```
wschin: Please add doc string and describe what types of the inputs. For metadata_props, you can add a reference to the ONNX page which defines allowed key-value pairs. It'd be nice if you can just copy one or two pairs from that page and then use them as an example here.

tests/coreml/test_AllNeuralNetworkConverters.py
```diff
@@ -1,8 +1,11 @@
+from distutils.version import StrictVersion
```
wschin: Please move this to line 7, right below `from coremltools.proto.FeatureTypes_pb2 import ImageFeatureType`. Thanks.

tiagoshibata: Will do. One question, do you have a reference for the order I should follow? You've asked me to move a few and I'm doing it blindly, but I'm a bit lost on the reasoning behind it. I'm following the [PEP 8 recommended ordering](https://www.python.org/dev/peps/pep-0008/#imports):

```
Imports should be grouped in the following order:

1. Standard library imports.
2. Related third party imports.
3. Local application/library specific imports.
```

Which would place distutils (standard library) above coremltools (third party) instead of the other way around.

wschin: My feeling is that sorting them by alphabet makes a cleaner view of imported packages. It's a simple and easy-to-be-remembered rule to me. :)

onnxmltools/utils/metadata_props.py
```diff
@@ -0,0 +1,82 @@
...
+    Set input type denotation and dimension denotation.
+
+    Type denotation is a feature in ONNX 1.2.1 that let's the model specify the content of a tensor (e.g. IMAGE or AUDIO).
+    This information can be used by the backend. One example where it is useful is in images: Whenever data is bound to
+    a tensor with type denotation IMAGE, the backend can process the data (such as transforming the color space and
+    pixel format) based on model metadata properties.
+
+    :param onnx_model: ONNX model object
+    :param input_name: Name of input tensor to edit
+    :param denotation: Input type denotation (`documentation <https://github.com/onnx/onnx/blob/master/docs/TypeDenotation.md#type-denotation-definition>`_)
```
wschin: What's its type? To know that it's a string, one needs to read [this file](https://github.com/onnx/onnx/blob/7848f1e0414ba3b2e263609d93d46fd60790b2e9/onnx/onnx.proto#L402) and trace the code below. However, for a new user, it could be very painful to figure out what this attribute should look like. Could you please specify all input types and add an example per attribute for all user-facing functions? I know you have a test but users are unlikely going to read it. Many thanks.

tiagoshibata: It it better now?

wschin: Yes, so only one last comment left; please copy the metadata from duplicate to original (as described in my general comment in my last review).

tiagoshibata:
Can this be merged? We at WinML would be happy if it reached master. Thanks!

wschin:
Sorry that I didn't realize that you cannot write into this repo.

tiagoshibata:
Thanks! :smile:

PR #116, opened 2018-07-24
Float16 convert

stevenlix:


wschin:
@wenbingl , the old code directly accesses the underlying data structure. It'd be better to use official interface to modify/create elements in repeated field.

wenbingl:
@wschin, but it will make the converter depend on the onnx ir implementation, right? Previously, we'd like to decouple them so that __targeted_onnx_version parameter can work well among different onnx versions.

wschin:
@wenbingl , it's protobuf's interface function, not ONNX's. No worries.

PR #120, opened 2018-07-27
Add Global Pooling 3D support for Keras Converter

wenbingl:


wschin:
Is it possible to add a CNTK-based test now?

wenbingl:
For some reason, CNTK is not encouraged for this kind of test.
But we are verifying it in some ways. stay tuned.

wenbingl:
verified by some internal model test.

PR #121, opened 2018-07-30
Adding onnx polish_model

duli2012:
Now we call a function from the official ONNX repo to check/polish the ONNX model generated by the converter.

### DIFF
--- a/onnxmltools/convert/main.py
+++ b/onnxmltools/convert/main.py
@@ -19,6 +19,7 @@ def convert_sklearn(model, name=None, initial_types=None, doc_string='',
     if targeted_onnx < StrictVersion('1.2'):
         return onnx_model
     else:
+        import onnx.utils
         return onnx.utils.polish_model(onnx_model)
 
 def convert_coreml(model, name=None, initial_types=None, doc_string='',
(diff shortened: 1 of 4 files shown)


onnxmltools/convert/main.py
```diff
@@ -6,33 +6,43 @@
...
 
     from .coreml.convert import convert
-    return convert(model, name, initial_types,
+    onnx_model = convert(model, name, initial_types,
                    doc_string, targeted_onnx, custom_conversion_functions, custom_shape_calculators)
-
+    if targeted_onnx < StrictVersion('1.2'):
+        return onnx_model
+    else:
+        return onnx.utils.polish_model(onnx_model)
```
wenbingl: Did you try some real model to see if the generated model can really satisfy the onnx checker?

wschin: We need to test this with all our models!

duli2012: For SKlearn models, we got onnx checker failer: "No opset import for domain "

wschin: This means we cannot call this function because ONNX-ML is not supported generally.

PR #127, opened 2018-08-09
check model's input and output naming convention

stevenlix:


onnxmltools/convert/common/_topology.py
```diff
@@ -648,6 +649,19 @@ def convert_topology(topology, model_name, doc_string, targeted_onnx):
                 else:
                     other_outputs[variable.raw_name] = variable
 
+    # Check input and output naming convention
+    for name in topology.raw_model.input_names:
+        input_name = name.replace('_', '')
+        if input_name[0].isdigit() or (not input_name.isalnum()):
```
wenbingl: is underline included in isalnum?

onnxmltools/convert/common/_topology.py
```diff
@@ -648,6 +649,19 @@ def convert_topology(topology, model_name, doc_string, targeted_onnx):
                 else:
                     other_outputs[variable.raw_name] = variable
 
+    # Check input and output naming convention
+    for name in topology.raw_model.input_names:
```
wenbingl: Do every model (CoreML, keras, scikit-learn) have this property - input_names?

wschin: `raw_model` is a wrapper class over the targeted source model. For Core ML model, its type is `CoremlModelContainer` defined [here](https://github.com/onnx/onnxmltools/blob/727cc8fadfbc17790978bb6c66db461ff62c764d/onnxmltools/convert/common/_container.py#L43). Similarly, we have `SklearnModelContainer` and `KerasModelContainer` for the other two frameworks. Please see [`RawModelContainer`](https://github.com/onnx/onnxmltools/blob/727cc8fadfbc17790978bb6c66db461ff62c764d/onnxmltools/convert/common/_container.py#L13) for their common behavior.

onnxmltools/convert/common/_topology.py
wschin: @stevenlix , why do we replace `_` with space? It's a C-stype name.

wenbingl:
The agreement regarding to C syntax name from yesterday meeting is to only report a warning if the  input/output names don't comply C syntax.

onnxmltools/convert/common/_topology.py
stevenlix: There is no change on any names. Here is to check whether the names are compliant with ONNX spec. If it's not, a warning is generated.

wschin: I got your point. So could you please merge the two new loops with the following ones (removing `break` so that all names can be examined)?

stevenlix: I think there is no need to generate warning for each name, so I just generate one warning as long as I detect an incompliant name. Users should check the model input/output names once they see the warning.

stevenlix: most likely, names are either all compliant or all incompliant

wschin: But the names are generated by us..

stevenlix: The input and output names are coming from original model that to be converted, aren't they?

wenbingl: From the end user view, it would be better to print more names which are incompliant.

onnxmltools/convert/common/_topology.py
```diff
@@ -649,17 +650,31 @@ def convert_topology(topology, model_name, doc_string, targeted_onnx):
...
+    invalid_name = []
     for name in topology.raw_model.input_names:
+        # Check input naming convention
+        input_name = name.replace('_', '')
+        if input_name and (input_name[0].isdigit() or (not input_name.isalnum())):
+            invalid_name.append(name)
         if name in tensor_inputs:
             container.add_input(tensor_inputs[name])
+    if invalid_name:
+        warnings.warn('The input names are not compliant with ONNX naming convention: %s' % invalid_name)
```
wschin: `The input` ---> `Some input`

onnxmltools/convert/common/_topology.py
```diff
@@ -649,17 +650,31 @@ def convert_topology(topology, model_name, doc_string, targeted_onnx):
...
+    invalid_name = []
     for name in topology.raw_model.output_names:
+        # Check output naming convention
+        output_name = name.replace('_', '')
+        if output_name and (output_name[0].isdigit() or (not output_name.isalnum())):
+            invalid_name.append(name)
         if name in tensor_outputs:
             container.add_output(tensor_outputs[name])
+    if invalid_name:
+        warnings.warn('The output names are not compliant with ONNX naming convention: %s' % invalid_name)
```
wschin: `The output` --> `Some output`

PR #128, opened 2018-08-14
change class lable string encoding as 'utf-8'.

wenbingl:


wenbingl:
that's intentional, since onnx spec only states the string tensor is encoded as 'utf-8'

wschin:
That string is appended to a list of strings and finally becomes a single [attribute](https://github.com/onnx/onnx/blob/master/docs/Operators.md#attributes-22).

wenbingl:
@wschin through very unlikely any issue on encoding activation string with ascii, it will be better to unify all the converter string into 'utf-8'.

PR #134, opened 2018-09-03
remove unnecessary print, add quote around filenames in some places, update unit test for scikit-learn 0.20

xadupre:
Printing should be removed or available with a condition such as verbose=True, this PR removes two print.


xadupre:
I looked for substring "print(" and remove what was not part of a command line.

onnxmltools/convert/common/utils.py
```diff
@@ -284,3 +284,18 @@ def check_input_and_output_types(operator, good_input_types=None, good_output_ty
                 raise RuntimeError('Operator %s (type: %s) got an output %s with a wrong type %s. Only %s are allowed' \
                                    % (operator.full_name, operator.type, variable.full_name, type(variable.type),
                                       good_output_types))
+
+
+def compare_strict_version(v1, v2):
+    if v1 is None:
```
wenbingl: It will easily confuse others when v1 is None, since it is not intuitive to define the behavior to compare a version with None.

tests/coreml/test_cml_SupportVectorClassifierConverter.py
```diff
@@ -47,7 +47,7 @@ def validate_zipmap(self, model):
         self.assertTrue('classProbability' in node.output)
 
     def test_support_vector_classifier_binary_no_prob(self):
-        svm = self._fit_binary_classification(SVC())
+        svm = self._fit_binary_classification(SVC(gamma=0.5))
```
wenbingl: Thanks for the fixing.
Should we add a partial function to wrap SVC(gamma=0.5) and add some comments about why gamma is 0.5?

onnxmltools/convert/common/utils.py
xadupre: I documented the function which now says:

The function always returns 1 if v1 is None, it is used to set the targeted version to a development version of ONNX.

tests/coreml/test_cml_SupportVectorClassifierConverter.py
xadupre: The default value changed in 0.20, 'auto' was deprecated. The current default value is 'auto_deprecated' and will be replaced by 'scale' later on. This should be revisited when sklearn 0.22 goes out. 0.20 introduced a couple of changes with default values.

Issue #138, opened 2018-09-26
Are the exported models intended to work with Windows.AI.MachineLearning?


pinkwerks:
Hi, I'm attempting to load model3.onnx (attached):
using Windows.AI.MachineLearning;

I get this error:
_model = await LearningModel.LoadFromStorageFileAsync(file);
Output:
Exception thrown: 'System.Runtime.InteropServices.COMException' in System.Private.CoreLib.dll
WinRT information: [TypeInferenceError] Invalid attribute perm {0, 3, 1, 2}, input shape = {1, 300}
Exception thrown: 'System.Runtime.InteropServices.COMException' in System.Private.CoreLib.dll
WinRT information: [TypeInferenceError] Invalid attribute perm {0, 3, 1, 2}, input shape = {1, 300}
An exception of type 'System.Runtime.InteropServices.COMException' occurred in System.Private.CoreLib.dll but was not handled in user code
WinRT information: [TypeInferenceError] Invalid attribute perm {0, 3, 1, 2}, input shape = {1, 300}
Unspecified error
[TypeInferenceError] Invalid attribute perm {0, 3, 1, 2}, input shape = {1, 300}

I'm not sure where to begin? I have another .onnx model file that loads okay with the same code. (also attached)
[models.zip](https://github.com/onnx/onnxmltools/files/2421210/models.zip)

pinkwerks:
To clarify, I wasn't sure if the file was invalid. Nutron was able to view them both so.ill assume this is a loader problem with windows.ai.

youngkim93:
Hi, thank you reporting your problem. In general, the fact that you can view the model in Netron does not imply that it can run on Windows.AI.Machinelearning platform. In the future, we will be providing a tool called winmltools that could potentially provide more information about validity of onnx model for Windows platform.

Having said that, we would still love to help this scenario! Were you able to solve this problem yet?

youngkim93:
It looks to me like the model does not have a right permutation attribute for transpose operator. Input Shape of transpose is 2 dimensions while permutation is specifying 4 dimensions. Is this model coming from the onnxmltool?

pinkwerks:
Hi, thanks for the help!

This model was created in Keras. Then exported using onnxmltools. I'm unclear if this is a translation problem, or the Windows.AI API is having an issue parsing something valid. FWIW the model "just worked" in keras. This is my model in keras and how I'm exporting it.

```
def baseline_model(num_classes=12, num_dim=33):
    # create model
[...]

PR #141, opened 2018-10-03
Fix batch normalization

youngkim93:
This change is to fix the case when batch normalization is applied to non 4-D dimensional tensors.


youngkim93:
#138

PR #146, opened 2018-10-09
Plumb default_batch_size for convert_* to all models

Oewyn:
Added the capability to set default_batch_size on convert_* calls. This is much easier than trying to use initalize_types to try and do the same thing.


Oewyn:
Closing for now

PR #147, opened 2018-10-09
Modifying operator_converters for keras to improve onnx node naming

Oewyn:
I found it challenging to do any sort of comparison between an onnx converted graph and the original keras graph because almost all of the names didn't match up. This PR is an attempt to address that. I have not modified all the keras operator_converters, but I did hit most of them (the ones remaining are namely RNN related -- Bidirectional, Embed, GRU, LSTM, SimpleRNN)

Any feedback would be helpful, I believe i got all the scope.get_unique_operator_name(...) vs scope.get_unique_variable_name(...) separated correctly.

### DIFF
--- a/onnxmltools/convert/keras/operator_converters/AdvancedActivation.py
+++ b/onnxmltools/convert/keras/operator_converters/AdvancedActivation.py
@@ -11,20 +11,21 @@ from ...common._registration import register_converter
 
 def convert_keras_advanced_activation(scope, operator, container):
     op = operator.raw_operator
+    keras_name = scope.get_unique_operator_name(operator.raw_operator.name)
     if isinstance(op, activations.LeakyReLU):
         alpha = op.get_config()['alpha']
         apply_leaky_relu(scope, operator.input_full_names[0], operator.output_full_names[0], container,
-                         operator_name=operator.full_name, alpha=alpha)
+                         operator_name=keras_name, alpha=alpha)
     elif isinstance(op, activations.ELU):
         alpha = op.get_config()['alpha']
         apply_elu(scope, operator.input_full_names[0], operator.output_full_names[0], container,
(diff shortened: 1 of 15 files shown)


Oewyn:
Yes, I've made sure that this change doesn't introduce any illegal characters.  Everything is run through scope.get_unique_* which checks and modifies keras supplied names and replaces illegal characters with '_'

onnxmltools/convert/keras/operator_converters/AdvancedActivation.py
```diff
@@ -11,20 +11,21 @@
 
 def convert_keras_advanced_activation(scope, operator, container):
     op = operator.raw_operator
+    keras_name = scope.get_unique_operator_name(operator.raw_operator.name)
     if isinstance(op, activations.LeakyReLU):
         alpha = op.get_config()['alpha']
         apply_leaky_relu(scope, operator.input_full_names[0], operator.output_full_names[0], container,
-                         operator_name=operator.full_name, alpha=alpha)
+                         operator_name=keras_name, alpha=alpha)
```
wenbingl: I didn't find the place where operator_name was referenced. Did I miss anything?

wenbingl:
Please add an unit test to verify these name changing really works.

onnxmltools/convert/keras/operator_converters/AdvancedActivation.py
Oewyn: operator_name is a named parameter to the apply_leaky_relu function. All apply_* functions have this parameter that allows the caller to set the name of the node container that is created.

Oewyn:
Thanks for the feedback, I should be able to get to the unit_tests early next week.

Oewyn:
Closing for now

PR #152, opened 2018-10-18
LightGBM Tests

little-eyes:
* binary classification convert
* multiclass classification convert
* regression convert


xadupre:
What is the difference with the tests implemented in https://github.com/onnx/onnxmltools/blob/master/tests/sklearn/test_TreeEnsembleConverters.py?

little-eyes:
> > What is the difference with the tests implemented in https://github.com/onnx/onnxmltools/blob/master/tests/sklearn/test_TreeEnsembleConverters.py?
> 
> Ah, I missed it. Also the multiclass LightGBM classification seems to have some issues. When I try to use it, i encountered into problems, that's why I want to add the multiclass test. Since I can't find a LightGBM test file, then I added one.

So I changed the PR to just add the multiclass tests. Actually, the problem I saw does not repro anymore..

wenbingl:
Any update of this PR, do we still need it?

xadupre:
It should be not needed anymore.

PR #161, opened 2018-10-31
Fixes #151, BernouilliNB converter

xadupre:

PR #162, opened 2018-11-01
Prroy/fix nb bugs

prabhat00155:
Fixed Bernoulli NB bug


xadupre:
I suggested a fix in https://github.com/onnx/onnxmltools/pull/161, I think yours is better.

PR #163, opened 2018-11-05
[WIP-do not merge] Opset 8 / Onnx 1.3 Updates

vinitra:
- checking for opset version instead of onnx version based on the [versioning list](https://github.com/onnx/onnx/blob/master/docs/Versioning.md#released-versions)
- Supporting numpy style broadcasting for max, min, mean, and sum
- (PENDING) add test support for broadcasting in CoreML converter


wenbingl:
Need make sure all your git commits with the same author name and email, otherwise there is a CLA error which will block the merge.

PR #166, opened 2018-11-07
ONNX 1.3 version update (Opset 8) for common functions

vinitra:
- Uses API from #157 to update common functions to check for opset number instead of ONNX version number (see [opset to version mapping](https://github.com/onnx/onnx/blob/master/docs/Versioning.md#released-versions))


vinitra:
@wenbingl

wschin:
@vinitra, you also need to `grep "add_node"` to double-check every use of `add_node(…)`. Otherwise, something like [this](https://github.com/onnx/onnxmltools/blob/master/onnxmltools/convert/coreml/operator_converters/neural_network/Pool.py#L317) will be missed. Please search for `targeted_onnx_version` in the whole codebase. Thanks a lot.

wenbingl:
@wschin,good reminder. Vinitra will have more PRs for that.

PR #167, opened 2018-11-07
Plumb default_batch_size for convert_* to keras/coreml

Oewyn:
Adding the capability to set default_batch_size on convert_keras and convert_coreml calls. This is much easier than trying to use initialize_types to try and do the same thing. As asked in a previous issue, I have gone through the keras code and concluded that indeed for Keras, the first dimension is always the batch dimension. Therefore using the same logic for coreml which has the same location for the batch dimension should be okay. The clearest example of such is the following reference code for keras reference implementation for batch_flatten backend function:

def batch_flatten(x):
    return np.reshape(x, (x.shape[0], -1))

### DIFF
--- a/onnxmltools/convert/keras/_parse.py
+++ b/onnxmltools/convert/keras/_parse.py
@@ -75,11 +75,13 @@ def determine_tensor_type(tensor, default_batch_size, keras_shape=None):
         raise ValueError('Unable to find out a correct type for tensor %s' % tensor)
 
 
-def parse_keras(model, initial_types=None, target_opset=None, targeted_onnx=onnx.__version__, custom_conversion_functions=None, custom_shape_calculators=None):
+def parse_keras(model, default_batch_size=1, initial_types=None, target_opset=None, targeted_onnx=onnx.__version__,
+                custom_conversion_functions=None, custom_shape_calculators=None):
     '''
     The main parsing function of Keras Model and Sequential objects.
 
     :param model: A Keras Model or Sequential object
+    :param default_batch_size: default batch size of produced ONNX model
     :param initial_types: A list providing some types for some root variables. Each element is a tuple of a variable
     name and a type defined in data_types.py.
     :param target_opset: number, for example, 7 for ONNX 1.2, and 8 for ONNX 1.3.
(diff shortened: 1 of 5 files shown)


wenbingl:
Thanks for the idea, if default_batch_size is adopted, we need add a check for initial_type and default_batch_size to avoid conflicts. would you like to add this  check as well?

Oewyn:
@wenbingl I can certainly take a look. I would need more information about what conflicts you wish to avoid. Do you mean a conflict where the user set default_batch_size to 2 and then set an input using initial_types to some shape that didn't have a '2' in the first dimension?

wenbingl:
> @wenbingl I can certainly take a look. I would need more information about what conflicts you wish to avoid. Do you mean a conflict where the user set default_batch_size to 2 and then set an input using initial_types to some shape that didn't have a '2' in the first dimension?

Yes, exactly! We should always have only one recommendation for the end user to change the default batch size.

Oewyn:
Okay, I can change the recommendation in the doc string to use default_batch_size rather than provide both options (initial_types/default_batch_size) but I will need a little more clarification on where to do any sort of initial_types vs default_batch_size checks. 

Inside the __init__ for the Topology class the default_batch_size is set to 1. Therefore, I'm not sure how it could check if default batch_size was set to something and initial_types was also set.

Are you proposing that default_batch_size should be None by default, and inside Topology.__init__() check if both default_batch_size is not None and initial_types is not None throw an error?

Then (presumably) after that, check if default_batch_size is None, set it to 1?

If that's not what you're thinking please give me additional guidance.

Thanks!

wenbingl:
Actually, the existing implementation is a little fishy, the batch_size could be overwritten by initial_types. Your idea about default_batch_size=None is doable, and add a warning about the potential conflicts.

Oewyn:
Updates made to add a warning and set default_batch_size to None

Oewyn:
I'm looking at the failing tests, and I'm trying to understand the intent of the test_image_input_type_converter function under tests/coreml/test_AllNeuralNetworkConverters.py -- It is making sure the first dimension is None instead of the default batch size of 1. This is related to the code in: convert/coreml/_parse.py:

```
# Determine the batch size for parsing CoreML model's input and output features. Note that batch size is always
# missing in all CoreML models.
default_batch_size = 1 if model.WhichOneof('Type') not in \
                          ['neuralNetworkClassifier', 'neuralNetworkRegressor', 'neuralNetwork'] else 'None'
```

This logic which I removed is in conflict with the default_batch_size/initial_types check you wanted to add @wenbingl . Is it okay if i modify the test to expect a '1' in the batch dimension?

wenbingl:
string 'None' in ONNX means there is dynamic dimension, and its value will be assigned after the input has decided. I just noticed in CoreML the batch size is always dynamic, why do we need the default batch size for it?

wschin:
@Oewyn, default batch size must be unknown to Core ML neural networks. For example, there is a [Exermote model](https://github.com/likedan/Awesome-CoreML-Models) where batch size is not specified due to Core ML spec. However, if you try to feed a input with batch smaller than 3, it will fail.

wenbingl:
Given the CoreML actually only take the batch size as an variable of input, it would be unnecessary to add a parameter to specify the default batch size. let's just restrict the change for keras converter only.

Oewyn:
Sounds good, I will work on updating the PR and have the changes posted by tomorrow.

Oewyn:
Updated to only impact Keras conversions, also fixed (how i believe it should be fixed) the issue with DepthwiseConv2D and an older keras version. I noticed that #173 also attempts to address this. But I've verified my changes actually work when calling convert_keras(...) on keras version 2.1.2. I'm not sure how just catching the exception upon import will work for cases where DepthwiseConv2D is actually used in code, e.g. keras/operator_converters/Conv.py:44

```if isinstance(op, DepthwiseConv2D):```

PR #168, opened 2018-11-07
Add an onnx optimizer and NCHW input conversion support.

wenbingl:


tests/end2end/test_keras_converter.py
```diff
@@ -19,8 +22,8 @@ def __init__(self, alpha=1.0, beta=1.0, **kwargs):
     def build(self, input_shape):
         super(ScaledTanh, self).build(input_shape)
 
-    def call(self, x):
-        return self.alpha * K.tanh(self.beta * x)
+    def call(self, inputs, **kwargs):
```
xadupre: **kwargs: is it necessary?

tests/end2end/test_keras_converter.py
```diff
@@ -44,15 +47,41 @@ def test_custom_op(self):
...
+        converted_model = onnxmltools.convert_keras(model, custom_conversion_functions={ScaledTanh: custom_activation})
+
         actual = model.predict(x)
         self.assertIsNotNone(actual)
 
-        converted_model = onnxmltools.convert_keras(model, custom_conversion_functions={ScaledTanh: custom_activation})
         self.assertIsNotNone(converted_model)
         # to check the model, you can print(str(converted_model))
-        dump_data_and_model(x.astype(np.float32), model, converted_model, basename="KerasCustomOp-Out0",
+        dump_data_and_model(x.astype(np.float32), model, converted_model, basename="Keras_custom_op-Out0",
```
xadupre: Why rename it?

tests/end2end/test_keras_converter.py
```diff
@@ -44,15 +47,41 @@ def test_custom_op(self):
...
+        model.add(MaxPooling2D((2, 2), strides=(2, 2), data_format='channels_last'))
+
+        model.compile(optimizer='sgd', loss='mse')
+        converted_model = onnxmltools.convert_keras(model, channel_first_inputs=[model.inputs[0].name])
+
+        expected = model.predict(x)
+        self.assertIsNotNone(expected)
+        self.assertIsNotNone(converted_model)
+
+        try:
```
xadupre: Why not using dump_data_and_model?

tests/end2end/test_keras_converter.py
wenbingl: compliant to keras interface to kill PEP warning.

tests/end2end/test_keras_converter.py
wenbingl: The inputs between keras and onnx model are different, which hasn't been supported by dump_data_and_model.

xadupre: Why not adding an extra parameter to support as transform (input_transform=lambda input: input.transpose( …)?

onnxmltools/convert/keras/convert.py
```diff
@@ -31,19 +31,22 @@ def convert(model, name=None, initial_types=None, doc_string='', target_opset=No
     and a type defined in data_types.py.
     :param doc_string: A string attached onto the produced ONNX model
     :param target_opset: number, for example, 7 for ONNX 1.2, and 8 for ONNX 1.3.
+    :param channel_first_inputs, enforce the input format of the converted model is channel_first, i.e. NCHW
```
wschin: Would this be better?
```suggestion
    :param channel_first_inputs: specifies names of 4-D inputs which are forced to be in channel-first format (i.e., NCHW) in the converted model. It's a list of string; for example, ['input1', 'input2'].
```

wschin:
My feeling is that you will end up with another `Topology` or `IR` for more sophisticated optimizations. Is it possible to reuse existing components?

PR #174, opened 2018-11-13
Stable

wenbingl:
Re-enable keras unit tests on the latest keras and coremltools.

### DIFF
--- a/onnxmltools/convert/common/_topology.py
+++ b/onnxmltools/convert/common/_topology.py
@@ -683,7 +683,8 @@ def convert_topology(topology, model_name, doc_string, target_opset, targeted_on
             invalid_name.append(name)
         if name in tensor_inputs:
             onnx_input = tensor_inputs[name]  # type: Variable
-            if name in channel_first_inputs:
+            if name in channel_first_inputs or \
+                    (name.endswith(':0') and name[:-2] in channel_first_inputs):
                 nhwc_inputs.append(onnx_input.full_name)
                 s = onnx_input.type.shape
                 onnx_input.type.shape = [s[0], s[3], s[1], s[2]]
--- a/onnxmltools/utils/tests_dl_helper.py
+++ b/onnxmltools/utils/tests_dl_helper.py
@@ -6,40 +6,67 @@
 """
 Helpers to help testing deep learning converted models.
 """
+import six
 import numpy as np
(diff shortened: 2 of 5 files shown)


onnxmltools/convert/common/_topology.py
```diff
@@ -683,7 +683,8 @@ def convert_topology(topology, model_name, doc_string, target_opset, targeted_on
             invalid_name.append(name)
         if name in tensor_inputs:
             onnx_input = tensor_inputs[name]  # type: Variable
-            if name in channel_first_inputs:
+            if name in channel_first_inputs or \
```
wschin: Hummm, it looks not good. Why do you need extra conditions?

onnxmltools/utils/tests_dl_helper.py
```diff
@@ -6,42 +6,69 @@
 """
 Helpers to help testing deep learning converted models.
 """
+import six
```
wschin: Please swap `six` and `numpy as np`.

tests/end2end/test_keras_converter.py
```diff
@@ -78,7 +78,7 @@ def test_channel_last(self):
             sess = onnxruntime.InferenceSession(converted_model.SerializeToString())
             actual = sess.run([], {sess.get_inputs()[0].name:
                                          np.transpose(x.astype(np.float32), [0, 3, 1, 2])})
-            self.assertTrue(np.allclose(expected, actual))
+            self.assertTrue(np.allclose(expected, actual, rtol=1.e-3))
```
wschin: The tolerance is relative large given the tiny size of the model.

onnxmltools/convert/common/_topology.py
wenbingl: to support the keras input name like 'input_1', instead of 'input_1:0'.

onnxmltools/utils/tests_dl_helper.py
wenbingl: Two different imports, and do you think it looks nice from short to the long.

tests/end2end/test_keras_converter.py
wenbingl: it's due to some random failure in travis test machine. Haven't figure out why.

PR #178, opened 2018-11-20
Add CountVectorizer (runtime not tested)

xadupre:


xadupre:
Extra spaces were intentional. Probably one of my editors with the wrong settings. I'll check that.

wenbingl:
The existing code are all using CR/LF, instead of LF only(Unix style), we should be consistent on it.

PR #183, opened 2018-11-22
Prroy/converter feature union

prabhat00155:
Feature union converter


onnxmltools/convert/sklearn/operator_converters/FeatureUnion.py
```diff
@@ -0,0 +1,38 @@
...
+        self.inputs = inputs
+        self.input_full_names = input_full_names
+        self.type = op_type
+
+def convert_sklearn_feature_union(scope, operator, container):
+    op = operator.raw_operator
+    transform_result_name = []
+    supported_transformer_list = [PCA, TruncatedSVD]
+    for name, transform in op.transformer_list:
+        if type(transform) not in supported_transformer_list:
```
xadupre: Why do we need to limite to a subset of transform?

onnxmltools/convert/sklearn/operator_converters/SVD.py
```diff
@@ -9,7 +9,7 @@
 from ...common._registration import register_converter
 
 
-def convert_truncated_svd(scope, operator, container):
+def convert_truncated_svd(scope, operator, container, flag=False):
```
xadupre: Why creating a special case for FeatureUnion?

prabhat00155: FeatureUnion would in turn invoke these converters and they need to return back the transformed tensor names (hence flag) back to the FeatureUnion converter where they can be concatenated and the output could be populated.

xadupre: I would prefer not to change the function signature but to create a new operator which overwrites inputs and outputs and then call the converter without any change. The flag parameter does not scale over all converters unless we add it anywhere. The duplication of code is ok for such simple converter but it might be difficult to maintain for bigger converters.

onnxmltools/convert/sklearn/operator_converters/FeatureUnion.py
prabhat00155: I'll need to change other transformers as well. I wanted get the changes to these two transformers reviewed first and then modify others.

onnxmltools/convert/sklearn/operator_converters/SVD.py
wenbingl: Agree with @xadupre.
Even there will be a large code dup, you can refactor it as a common function.

prabhat00155: Fixed.

wschin:
Why do you have so many commits?

onnxmltools/convert/sklearn/operator_converters/FeatureUnion.py
```diff
@@ -0,0 +1,44 @@
...
+
+class Oper:
+    def __init__(self, model, inputs, op_type, input_full_names, output_full_name):
+        self.raw_operator = model
+        self.inputs = inputs
+        self.input_full_names = input_full_names
+        self.type = op_type
+        self.outputs = [Output(output_full_name)]
+
+def convert_sklearn_feature_union(scope, operator, container):
```
wschin: This converter needs document to explain the overall idea. Please also add inline comments for non-trivial details.

onnxmltools/convert/sklearn/operator_converters/SVD.py
```diff
@@ -22,7 +22,7 @@ def convert_truncated_svd(scope, operator, container):
     if operator.type == 'SklearnTruncatedSVD': # TruncatedSVD 
         # Create the major operator, a matrix multiplication.
         container.add_node('MatMul', [operator.inputs[0].full_name, transform_matrix_name],
-                           operator.outputs[0].full_name, name=operator.full_name)
+                           operator.outputs[0].full_name, name=scope.get_unique_variable_name('MatMul1'))
```
wschin: Why do we start using `MatMul1`? Can't it be `MatMul`?

onnxmltools/convert/sklearn/operator_converters/__init__.py
```diff
@@ -8,6 +8,7 @@
 from . import Binarizer
 from . import DictVectorizer
 from . import DecisionTree
+from . import FeatureUnion
```
wschin: Nice position!

onnxmltools/convert/sklearn/operator_converters/FeatureUnion.py
```diff
@@ -0,0 +1,44 @@
...
+    op = operator.raw_operator
+    transform_result_name = []
+    supported_transformer_list = [PCA, TruncatedSVD]
+    for name, transform in op.transformer_list:
+        if type(transform) not in supported_transformer_list:
+            raise NotImplementedError
+        op_type = sklearn_operator_name_map[type(transform)]
+        result_name = scope.get_unique_variable_name('result')
```
wschin: Looks like you didn't edit the parser but put all logics into `FeatureUnion`s converter. I perfer to create of ONNXMLTOOLS Operator during the parsing stage. This design somehow violates my original idea: parsing stage produces IR objects and those IR objects are converted at conversion stage. Let me try to add some details about a design in my mind.
1. Go to [this function](https://github.com/onnx/onnxmltools/blob/533c877194ed2c168f61219a03a4a094f5e1cd80/onnxmltools/convert/sklearn/_parse.py#L173) and add a new `_parse_sklearn_feature_union(scope, model, inputs)`
2. An implementation of `_parse_sklearn_feature_union(scope, model, inputs)`
```python
# scope is Scope in our IR
# model is FeatureUnion
# inputs is a string list with a single element
_parse_sklearn_feature_union(scope, model, inputs):
    # Output variable name of each transform. It's a list of string.
    transformed_result_names = []
    # Encode each transform as our IR objects
    for transform in model.transformer_list:
        transformed_result_names.add(_parse_sklearn_simple_model(scope, transform, inputs)
    # Create a Concat ONNX node
    concat_operator = scope.declare_local_operator('SklearnConcat')
    concat_operator.inputs = transformed_result_names

    # Declare output name of scikit-learn FeatureUnion
    union_name = scope.declare_local_variable('union')
    concat_operator.outputs.add(union_name)

    return union_name
```
3. Create a converter for `SklearnConcat` to `onnxmltools/convert/sklearn/operator_converters/SklearnConcat.`
4. Create a shape calculator for `SklearnConcat`.

PR #186, opened 2018-11-27
fix the optimizer on some extreme test cases.

wenbingl:


onnxmltools/convert/common/optimizer.py
```diff
@@ -35,7 +35,8 @@ def is_transpose(self):
 
     @property
     def in_single_path(self):
-        return len(self.successor) <= 1
```
wschin: Do you mind to add a doc string for this function?

tests/utils/test_utils.py
```diff
@@ -106,16 +106,17 @@ def test_optimizer(self):
         nodes[3:] = [helper.make_node('Max', ['input1', 'identity2'], ['max0'])]
         nodes[4:] = [helper.make_node('Transpose', ['max0'], ['tranpose0'], perm=[0, 2, 3, 1])]
         nodes[5:] = [helper.make_node('Transpose', ['tranpose0'], ['tranpose1'], perm=(0, 3, 1, 2))]
+        nodes[6:] = [helper.make_node('Relu', ['tranpose1'], ['output0'], perm=(0, 3, 1, 2))]
```
wschin: This is a very good change. Optimizer should not touch input/output at all.

onnxmltools/convert/common/optimizer.py
```diff
@@ -35,7 +35,8 @@ def is_transpose(self):
...
+        return len(self.successor) == 1 and not self.successor[0].in_or_out and \
+               len(self.precedence) == 1 and len(self.precedence[0].successor) <= 1
 
     @property
     def in_or_out(self):
```
wschin: I don't quite understand this attribute.

wenbingl: Optimization could not happen on fan in/out node.

onnxmltools/convert/common/optimizer.py
wenbingl: added

PR #192, opened 2018-12-05
Fixes #18, add converters for xgboost

### LINKED ISSUE #18: XGBoost support
see https://github.com/KOLANICH/pyxgboost for xgboost parser and how to use it.
Feel free to split that project into two: xgboost -> onnx and onnx -> python

xadupre:
Regressor is ok, classifiers still need to be fixed.


KOLANICH:
You may find [my parser for XGBoost binary format](https://gitlab.com/KOLANICH/pyxgboost) useful

xadupre:
Thanks for the offer, I saw you called xgboost executable and then parse the output to retrieve the coefficient. I chose a different option by using the API to jump the model into json format. It was introduced in XGBoost 0.7. I prefer to use this way even if it forces the user to use a recent version of xgboost. What do you think?

KOLANICH:
For storage I prefer to use the binary format because JSON one has larger overhead, but IDK what people prefer (except the obvious leader of pickle used by the one who don't care about security).

xadupre:
Just to clarify, I use JSON to retrieve the coefficients from the xgboost model but the model is then stored into ONNX format which is a binary format based on protobuf. I did not compare the sizes of both ONNX and xgboost format but they should be pretty similar.

KOLANICH:
>I use JSON to retrieve the coefficients from the xgboost model but the model is then stored into ONNX format which is a binary format based on protobuf.

I know, it is interchange format to import the model into GPU computation network frameworks.

I mean that it may be useful to have the converter workable without xgboost itself, since we read serialized models, we don't need xgboost library itself.

youngkim93:
Hi Xavier, I'm curious to know what new support we have in this PR from previous xgboost converter in winmltools. Thanks!

xadupre:
The converter is now tested against onnxruntime and follows the same API as every converter in onnxmltools. That's easier to maintain in this module.

youngkim93:
Yeah I agree that all the converter logic should live in onnxmltools. Thanks for doing this! The last converter that is still in winmltools is libsvm. Any plan to migrate that as well?

xadupre:
That's PR #184. I have a couple of PRs which are not merged yet :)

wenbingl:
Please note that the travis yaml was broken in this PR so the travis check was skipped actually.

youngkim93:
Can we update the readme.md to reflect dependencies on xgboost and libsvm? For libsvm, I think there is no distribution in pypi so it would be good to provide a way to install one from different sources (maybe https://www.lfd.uci.edu/~gohlke/pythonlibs/#libsvm)?

xadupre:
I updated readme.md to add xgboost and libsvm. I did not add a link on https://www.lfd.uci.edu/~gohlke/pythonlibs/#libsvm as it is built from its sources everytime it is tested (see appveyor.yml). Users can check how it is installed on Windows and Linux by looking at CI files.

PR #196, opened 2018-12-10
Support benchmarking when comparing the results between onnx and other models

xadupre:
To run it: python tests/main.py


tests/sklearn/test_SklearnNaiveBayesConverter.py
```diff
@@ -18,6 +18,12 @@ def _fit_model_binary_classification(self, model):
         model.fit(X, y)
         return model, X.astype(numpy.float32)
```
vinitra: This file no longer exists in the master branch due to the refactoring. You might want to pull from master, or consider moving these changes to the skl2onnx repo!

PR #217, opened 2019-01-23
Add the missing keras activation parameter in the conv layers.

wenbingl:


faruknane:
Hey, so if I use the previous versions, is it gonna do its job? Thank you

PR #222, opened 2019-01-29
Skl2onnx wrapper, deleting redundant sklearn converter/testing code

vinitra:
* including dependency on skl2onnx for onnxmltools 1.4
* creating wrapper for scikitlearn to onnx model conversion
* deleting sklearn converter files and text to remove redundant code


onnxmltools/convert/common/utils.py
```diff
@@ -20,6 +20,15 @@ def sklearn_installed():
     except ImportError:
         return False
 
+def skl2onnx_installed():
```
xadupre: If you take a strong dependency on skl2onnx, import skl2onnx should be enough (you Added skl2onnx into requirements.txt).

Maybe you can also remove types in data_types. Having the same classes implemented the same way in both packages might be confusing.

PR #226, opened 2019-02-01
Delete keras converter code from onnxmltools and add a wrapper for keras2onnx

jiafatom:


onnxmltools/convert/main.py
```diff
@@ -21,12 +21,11 @@ def convert_keras(model, name=None, initial_types=None, doc_string='',
                   target_opset=None, targeted_onnx=onnx.__version__,
                   channel_first_inputs=None, custom_conversion_functions=None, custom_shape_calculators=None,
```
wenbingl: Add a warning here to say the custom_conversion_XXX is not supported any more if this argument is not empty.

onnxmltools/convert/common/utils.py
```diff
@@ -41,12 +41,12 @@ def coreml_installed():
         return False
 
 
-def keras_installed():
+def ketone_installed():
```
vinitra: Did you get a chance to include ketone in the requirements.txt file?

wenbingl: we haven't publish the pip package yet, let's fix it after the package was released.

PR #227, opened 2019-02-01
Add a small tool to remove cast node from onnx model

wenbingl:


jiafatom:
onnxmltools-linux-conda-ci fails

wenbingl:
> why we remove the unit test test_onnx_model here?

Let's find more regular model for this test.

PR #230, opened 2019-02-06
Rename ketone to keras2onnx

jiafatom:
we need consume ketone package, so rename here after we did this for keras-onnx.


vinitra:
@jiafatom You might want to add it to the requirements.txt file as well!

jiafatom:
@vinitra , keras2onnx is not published yet; we will add this after it is published.

PR #238, opened 2019-02-12
Sparkml

jeffsaremi:
This is the initial version and experimental effort in converting Spark ML models to ONNX
Currently only 4 operators are supported:
- StringIndexer
- OneHotEncoderEstimator
- LogisticRegression
- VectorAssembler


requirements-dev.txt
```diff
@@ -5,11 +5,12 @@ tensorflow
...
 pandas
-pytest
 pytest-cov
 scikit-learn
 scipy
 svm
 wheel
 xgboost
 lightgbm
+pyspark==2.3.2
```
wenbingl: any reason we have to stick to this version?

onnxmltools/convert/sparkml/__init__.py
```diff
@@ -0,0 +1,8 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+from .convert import convert
+from .utils import *
```
wenbingl: can we avoid * import here? pylint always suggest we should not use import *

jeffsaremi: I'll change it and make it specific

requirements-dev.txt
jeffsaremi: version 2.4 does not work. Until they provide fixes for that yes we need to stick with this

PR #239, opened 2019-02-13
fix CI on windows + remove all references to keras

xadupre:


onnxmltools/convert/main.py
```diff
@@ -18,20 +18,6 @@ def convert_coreml(model, name=None, initial_types=None, doc_string='', target_o
                    custom_conversion_functions, custom_shape_calculators)
 
 
-def convert_keras(model, name=None, initial_types=None, doc_string='',
```
wenbingl: why is the keras converter wrapper removed?

.azure-pipelines/win32-conda-CI.yml
```diff
@@ -19,13 +19,11 @@ jobs:
       Python35:
         python.version: '3.5'
         ONNX_PATH: onnx==1.2.3
-        KERAS: keras==2.1.6
         COREML_PATH: https://github.com/apple/coremltools/archive/v2.0.zip
 
       Python36:
-        python.version: '3.6'
+        python.version: '3.6.7'
```
wenbingl: which converter is specific to this python version?

xadupre: I wanted to check if the issue with numpy could be solved without a previous python version. I'll remove it.

onnxmltools/convert/main.py
xadupre: keras moved to keras-onnx?

wenbingl: this is a wrapper function, just like what we did for skl2onnx

xadupre: I think we should not do that. It is confusing. It also suggests onnxmltools takes a dependency on keras-onnx and it should not.

wenbingl: The decision is onnxmltools will become a portal of the converter collections. To save the cost, some small converter is still staying in this git repro.
Any concerns of not doing this?

xadupre:
I'm not sure it is a good decision. Two functions have the same name, one in onnxmltools, one in keras-onnx, what happens if the function in keras-onnx changes, we need to update the documentation as well, the signature. It is difficult to maintain. If users want to convert a keras model, they should use keras-onnx and not a package which might come with extra dependencies. Plus, onnxmltools does not add anything to it, so to me, it is just duplicated code.

wenbingl:
I don't think the issue you mention really bother the end users. If the user only wants to use the keras converter, s/he can install keras2onnx only. Regarding to the document, it will only provide a link in onnxmltools and all keras related stuff will move to keras2onnx project.

 Two scenarios that we have to do it is 1) backward compatibility 2) The converter pipeline or other similar app would like to a center place to collect all converters, instead of collecting them by themselves.

xadupre:
skl2onnx already handles pipelines with custom converters. It is not perfect because we have duplicated code in https://github.com/onnx/onnxmltools/blob/master/onnxmltools/convert/common/data_types.py and this piece of code must be common to every converter.

Backward compatibility... you're right, I should have put a deprecated warning instead. But duplicated code is always difficult to maintain.

wenbingl:
We already has a work item to move the common folder into onnxtk so that every converter depends on it and share the data types. No one takes this work item yet.

PR #242, opened 2019-02-15
Fixes #233, improve error handling

### LINKED ISSUE #233: Unsufficient error handling in save_model
`onnxmltools.save_model` does not seem to raise exceptions on incorrect input.

**Example**

```py
import onnxmltools

model = None

onnxmltools.save_model(model, './estimator.onnx')
```

**Expected result**

A Python exception with a message mentioning that `model` is not a valid onnxml model.

**Actual result**
No python exception. A printed message,
```
Failed trying to save file './estimator.onnx'.
```

**Version**

onnxmltools version 1.3.1

vinitra:
- Adds more details when loading / saving model fails
- Raises error if model argument is not an ONNX model


onnxmltools/utils/main.py
```diff
@@ -34,7 +34,7 @@ def load_model(file_path):
         print("File '{0}' could not be opened.".format(file_path))
         return
     except:
-        print("Could not load protobuf file.")
+        print("Could not load protobuf file '{0}'.".format(file_path))
```
rth: Maybe raise the above 2 prints as exceptions as well?

onnxmltools/utils/main.py
```diff
@@ -81,11 +83,17 @@ def save_text(model, file_path):
...
     try:
         with open(file_path, "w") as f:
             f.write(str(model))
     except IOError:
-        print("Could not save file")
+        print("Unable to write file to path '{0}', check if you have permissions.".format(file_path))
+        return
+    except:
+        print("Failed trying to save file '{0}'.".format(file_path))
+        return
```
rth: Is this exception catching really necessary?

If something goes wrong e.g. `file_path` does not exist or `str(model)` raises an exception it's better IMO to get the actual exception than an obscure `Failed trying to save file '<file_name>'` that doesn't provide any useful information for bug reports..

vinitra: Thanks for the suggestions. I agree that it makes more sense not to catch a general except. See latest update!

vinitra: @wenbingl, is there a reason why the team chose to use print statements instead of exceptions?

wenbingl:
The initial thinking is that the converter don't abort the user whole conversion process if the error was encountered. If we want to raise exception we need to continue to raise the original one instead of creating a new one.

onnxmltools/utils/main.py
```diff
@@ -31,10 +31,10 @@ def load_model(file_path):
         with open(file_path, 'rb') as f:
             model.ParseFromString(f.read())
     except IOError:
-        print("File '{0}' could not be opened.".format(file_path))
+        raise Exception("File '{0}' could not be opened.".format(file_path))
```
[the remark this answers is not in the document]
vinitra: Right! Thanks for the catch. Latest update removes try-except blocks.

PR #251, opened 2019-03-01
CoreML Opset 9 Updates: Constant, Flatten, GEMM

vinitra:
- Constant (LoadConstant, TensorToLabel)
- Flatten (Flatten)
- GEMM (InnerProduct)

Other relevant operator updates (BatchNormalization, PRelu, Cast, Upsample) have already been included in the common files in PR #241


vinitra:
Decided not to update CoreML converter's [OneHotEncoder](https://github.com/onnx/onnxmltools/blob/master/onnxmltools/convert/coreml/operator_converters/OneHotEncoder.py) implementation with ONNX Opset 9's [OneHot](https://github.com/onnx/onnx/blob/master/docs/Operators.md#OneHot) operator since it's already implemented using [OneHotEncoder](https://github.com/onnx/onnx/blob/master/docs/Operators-ml.md#ai.onnx.ml.OneHotEncoder) from the ONNX ML domain.

onnxmltools/convert/coreml/operator_converters/TensorToLabel.py
```diff
@@ -66,7 +66,11 @@ def convert_tensor_to_label(scope, operator, container):
     label_loader_attrs = {'name': label_loader_name}
     label_buffer_name = scope.get_unique_variable_name('ClassLabels')
     label_loader_attrs['value'] = helper.make_tensor(label_buffer_name, label_type, [len(labels)], labels)
-    container.add_node('Constant', [], [label_buffer_name], **label_loader_attrs)
+    if container.target_opset < 9:
```
wenbingl: should we create a apply_XXX-like function for it?

vinitra: Makes sense! I'll add apply_constant so we don't have to repeat the logic in multiple places.

onnxmltools/convert/common/_apply_operation.py
```diff
@@ -162,28 +159,80 @@ def apply_concat(scope, input_names, output_name, container, operator_name=None,
 
     container.add_node('Concat', input_names, output_name, op_version=op_version, name=name, axis=axis)
 
+def apply_constant(scope, input_names=[], output_name, operator_name=None, value=0.0):
+    name = _create_name_or_use_existing_one(scope, 'Constant', operator_name)
+    attrs = {'name': name, 'value': value}
 
-def apply_clip(scope, input_name, output_name, container, operator_name=None, max=None, min=None):
-    name = _create_name_or_use_existing_one(scope, 'Clip', operator_name)
+    if container.target_opset < 9:
```
wenbingl: well, if we don't use any attr or schema from opset 9, should we just keep it as 1

vinitra: In opset 9, the constant op was expanded to work with all input/output types. In opset 1, constant only worked with float tensor input. Does it make sense to add a data type check in the opset 1 case?

PR #254, opened 2019-03-06
Updating ONNX version (v1.4.1) in CI builds

vinitra:
* As title, in conjunction with converter ONNX opset 9 updates
* Added test_utils reference output for save_txt changes to initializer data_type
* Added input change to upsample op


vinitra:
CI test build failure looks to be due to data_type=7 instead of data_type="INT64" in the save_txt representation of initializers for "ArrayFeatureExtractor"

Issue #259, opened 2019-03-11
Deprecation of experimental ops for CoreML conversion, Opset 10


vinitra:
Affine -> Mul+Add: #258 
Crop -> DynamicSlice: #249 
ImageScaler -> Mul+Add: #257 
ParametricSoftplus -> Log+Exp+Mul+Add: #260
ScaledTanh -> Mul+Tanh: #265 
~~ThresholdedRelu -> Greater+Mul+Add~~: Promoted to an ONNX op for opset 10 https://github.com/onnx/onnx/pull/1856

PR #265, opened 2019-03-19
Decompose ScaledTanh into ONNX operators

vinitra:
In preparation for experimental op deprecation in the next ONNX release (v1.5, opset 10). Last operator in #259.

### DIFF
--- a/onnxmltools/convert/common/_apply_operation.py
+++ b/onnxmltools/convert/common/_apply_operation.py
@@ -421,12 +421,16 @@ def apply_parametric_softplus(scope, input_name, output_name, container, operato
         apply_mul(scope, [fName, aName], output_name, container)
 
 
-def apply_scaled_tanh(scope, input_name, output_name, container, operator_name=None, alpha=[1.0], beta=[1.0]):
+def apply_scaled_tanh(scope, input_name, output_name, container, operator_name=None, alpha=None, beta=None):
+    if alpha == None:
+        alpha = [1.0]
+    if beta == None:
+        beta = [1.0]
+    if len(alpha) != 1 or len(beta) != 1:
+        raise ValueError('alpha and beta must be 1-element lists')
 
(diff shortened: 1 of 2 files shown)


onnxmltools/convert/common/_apply_operation.py
```diff
@@ -420,6 +420,39 @@ def apply_parametric_softplus(scope, input_name, output_name, container, operato
...
         apply_mul(scope, [fName, aName], output_name, container)
 
+
+def apply_scaled_tanh(scope, input_name, output_name, container, operator_name=None, alpha=[1.0], beta=[1.0]):
+
+    name = _create_name_or_use_existing_one(scope, 'ScaledTanh', operator_name)
+    if container.target_opset < 9:
+        if len(alpha) != 1 or len(beta) != 1:
+            raise ValueError('alpha and beta must be 1-element lists')
+        op_type = 'ScaledTanh'
```
wschin: `op_type` is not necessary. We can just do `container.add_node('ScaledTanh', input_name, output_name, **attrs)

vinitra: Thanks for the suggestion! Fixed in update.

onnxmltools/convert/common/_apply_operation.py
```diff
@@ -420,6 +420,39 @@ def apply_parametric_softplus(scope, input_name, output_name, container, operato
         # g = a * f
         apply_mul(scope, [fName, aName], output_name, container)
 
+
+def apply_scaled_tanh(scope, input_name, output_name, container, operator_name=None, alpha=[1.0], beta=[1.0]):
```
wschin: Please don't do `alpha=[1.0]`. Use `None` instead. Pycharm can tell why.

vinitra: Thanks for the catch!

PR #266, opened 2019-03-20
Separating initializers from container inputs in opset 9+ graph creation

vinitra:
As title. Relevant PR -- https://github.com/onnx/keras-onnx/pull/42.

### DIFF
--- a/onnxmltools/convert/common/_topology.py
+++ b/onnxmltools/convert/common/_topology.py
@@ -46,7 +46,7 @@ class Variable:
         Return a globally unique variable ID
         '''
         return self.onnx_name
-    
+
     def __str__(self):
         if self.raw_name != self.onnx_name:
             return "Var(name='{0}', onnx='{1}', type={2})".format(self.raw_name, self.onnx_name, self.type)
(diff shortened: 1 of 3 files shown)


onnxmltools/convert/common/_topology.py
```diff
@@ -726,26 +726,35 @@ def convert_topology(topology, model_name, doc_string, target_opset, targeted_on
...
-            continue
-
-        # Initializers are always tensors so we can just call make_tensor_value_info(...)
-        value_info = helper.make_tensor_value_info(tensor.name, tensor.data_type, tensor.dims)
-        extra_inputs.append(value_info)
+    # When calling ModelComponentContainer's add_initializer(...), nothing is added into the input list.
+    # However, for ONNX target opset < 9, initializers should also be model's (GraphProto) inputs.
+    # Thus, we create ValueInfoProto objects from initializers (type: TensorProto) directly and
+    # then add them into model's input list.
+    if container.target_opset < 9:
```
wenbingl: why not follow the changed was made on keras converter? This change actually breaks the optimize_onnx because it needs to know the initializer list.

vinitra: Just realized after talking to David! Pushed an update with the fix.

PR #268, opened 2019-03-23
Refactor onnxmltools common folder into onnxtk

vinitra:
- [x] Moving files from onnxmltools/convert/common -> onnxutils/onnxtk/common
- [x] Fixing dependencies, broken tests, imports, CI builds, updating folder structure.
- [x] ~~onnxtk renaming discussion~~ (postponed to a future PR, as this one is complicated enough as it is)
- [x] remove duplicated files -- i.e. optimizer & add_metadata_props
TODO: update onnxtk PyPi package + documentation

PR highlights: 
- move onnxmltools/convert/common folder files to onnxtk/common
- replace imports of onnxmltools.convert.common with onnxtk.common
- build onnxtk in CI builds
- remove onnxmltools optimizer, use onnxtk optimizer
- in onnxtk, import onnx_pb as onnx_proto directly instead of depending on the proto/* files in onnxmltools


onnxutils/onnxtk/common/utils.py
```diff
@@ -145,31 +145,31 @@ def get_producer():
     """
     Internal helper function to return the producer
     """
-    from ... import __producer__
+    from onnxmltools import __producer__
```
vinitra: Unsure if this is the right package to draw from, but when calling from the _init_.py file in onnxtk, the "domain" causes test failures due to the disparity between the location of the tests and the location of the helper functions. @wenbingl, thoughts?

vinitra: Resolved to using native package __producer__ with ".." in recent update.

.azure-pipelines/win32-conda-CI.yml
```diff
@@ -46,13 +46,20 @@ jobs:
       echo Test numpy installation... && python -c "import numpy"
       pip install %COREML_PATH% %ONNX_PATH%
       pip install -r requirements-dev.txt
+      pip install onnxruntime==0.1.4
       echo Test onnxruntime installation... && python -c "import onnxruntime"
```
vinitra: I didn't find an onnxruntime install in the Win32 CI build, although it checks for it in line 50, so I added it above. Is this assumption correct?

.azure-pipelines/linux-conda-CI.yml
```diff
@@ -44,11 +44,17 @@ jobs:
       git clone --recursive https://github.com/cjlin1/libsvm libsvm
       cd libsvm
       make lib
+      mkdir dependencies && cd dependencies
+      git clone https://github.com/vinitra/onnxmltools
+      cd onnxmltools/onnxutils
+      git fetch && git checkout refactor-common
+      pip install -e .
     displayName: 'Install dependencies'
```
vinitra: I've currently been installing onnxtk from the source of this PR in my fork as the dependencies have shifted. Once this has been checked into master, then it makes sense to install onnxtk from the source of the onnx/onnxmltools repo as well as publish a new version of the [onntk package to pypi](https://pypi.org/project/onnxtk/), ideally with more cohesive documentation.

onnxutils/onnxtk/metadata_props.py
```diff
@@ -0,0 +1,83 @@
...
+import onnx
+from onnx import onnx_pb as onnx_proto
+
+KNOWN_METADATA_PROPS = CaseInsensitiveDict({
+    'Image.BitmapPixelFormat': ['gray8', 'rgb8', 'bgr8', 'rgba8', 'bgra8'],
+    'Image.ColorSpaceGamma': ['linear', 'srgb'],
+    'Image.NominalPixelRange': ['nominalrange_0_255', 'normalized_0_1', 'normalized_1_1', 'nominalrange_16_235'],
+})
+
+
```
vinitra: I included the metadata_props.py file from the onnxmltools/utils as part of onnxtk since the dependency should be onnxtk -> onnxmltools, but not the other way around. _topology.py, in the common folder, depends on add_metadata_props.

.azure-pipelines/win32-conda-CI.yml
```diff
@@ -46,13 +46,18 @@ jobs:
       echo Test numpy installation... && python -c "import numpy"
       pip install %COREML_PATH% %ONNX_PATH%
       pip install -r requirements-dev.txt
-      echo Test onnxruntime installation... && python -c "import onnxruntime"
```
vinitra: I noticed there was no onnxruntime installation in the win32 build -- when 0.1.4 is added, one of the logistic regression spark ml tests fails. Without 0.1.4, no tests explicitly fail. I've informed @jeffsaremi regarding the log-reg test, but is the lack of onnxruntime in the windows build expected?

onnxmltools/convert/coreml/operator_converters/neural_network/Embed.py
```diff
@@ -7,8 +7,8 @@
 import numpy as np
 from .Scale import deduce_broadcast_axis_and_shape
 from .....proto import onnx_proto
-from ....common._apply_operation import apply_add, apply_reshape, apply_cast
-from ....common._registration import register_converter
+from onnxtk.common._apply_operation import apply_add, apply_reshape, apply_cast
```
wenbingl: As we discussed today, we should keep the common folder and in this folder we can import the types and functions from onnxtk, so we can avoid a lot of changes in the existing converter functions and had a better backward compatibility.

.azure-pipelines/win32-conda-CI.yml
vinitra: Will be addressed in separate PR.

.azure-pipelines/linux-conda-CI.yml
wenbingl: the onnxmltools local clone should already have any change in this PR? can't we just cd onnxmltools/onnxutils to install it? why do we have to install it from your fork?

vinitra: Ah, that makes a lot more sense! I didn't realize that the CI build creates a local clone as I didn't see it expressed in the yml files. Updated with a fix.

onnxmltools/convert/coreml/operator_converters/neural_network/Embed.py
vinitra: Updated in latest version!

PR #270, opened 2019-03-26
Support objective='reg:logistic' for XGBoost converters

xadupre:


tests/xgboost/test_xgboost_converters.py
```diff
@@ -49,6 +49,32 @@ def test_xgb_classifier_multi(self):
         self.assertTrue(conv_model is not None)
         dump_multiple_classification(xgb, allow_failure="StrictVersion(onnx.__version__) < StrictVersion('1.3.0')")
 
+    @unittest.skipIf(sys.version_info[0] == 2, reason="xgboost converted not tested on python 2")
```
prabhat00155: Is 'converted' a typo? It should either be converter or converted model. Same comment below.

xadupre: Fixed.

PR #275, opened 2019-04-08
Fix model save_text issue by json.dumps

jiafatom:


onnxmltools/utils/main.py
```diff
@@ -71,7 +72,7 @@ def save_text(model, file_path):
     if model is None or not isinstance(model, onnx_proto.ModelProto):
         raise ValueError("Model is not a valid ONNX model.")
     with open(file_path, "w") as f:
-        f.write(str(model))
+        f.write(json.dumps(str(model)))
```
wenbingl: What's the problem of the existing text output?

jiafatom: save_text does not generate json object.

jiafatom:
Synced offline, we are not going to support this.

PR #276, opened 2019-04-09
Incorporate data_type changes from skl2onnx

vinitra:
Include string representations of data types and a check to identify if ONNX was built with ONNX-ML flag.

Relevant changes originally made in https://github.com/onnx/sklearn-onnx/pull/93 and https://github.com/onnx/sklearn-onnx/pull/40.


onnxutils/onnxtk/common/data_types.py
```diff
@@ -111,6 +127,9 @@ def __init__(self, shape=None, doc_string=''):
     def _get_element_onnx_type(self):
         return onnx_proto.TensorProto.STRING
 
+    def __repr__(self):
+        return "StringTensorType(shape={0})".format(self.shape)
```
wenbingl: will it cause any exception if shape is None?

vinitra: No, it will just return: 'StringTensorType(shape=None)'

vinitra: Thanks for the review!

PR #278, opened 2019-04-09
Rename onnxtk to onnxconverter-common, implement API fixes

vinitra:
+ rename onnxtk package to onnxconverter-common
+ update in CI builds, onnxmltools wrappers, tests
+ remove common subfolder and place all files in onnxconverter-common root directory
+ adhere to style guidelines and rename public facing API files without starting _ (i.e. container instead of _container)
+ more appropriate name for _apply_operation -> onnx_ops.py


onnxutils/onnxconverter_common/utils.py
```diff
@@ -196,7 +196,7 @@ def is_string_type(item):
     return isinstance(item, types)
 
 
-def _check_has_attr(obj, attribute):
+def check_has_attr(obj, attribute):
```
wenbingl: well, it's not a good practice to rename a python function, which looks weird and unncessary.
Let's remove it.

vinitra: Got it! I'll replace it's use in the SVM converter with the native Python has_attr.

girdevstack:
For anyone that is happening to be looking for where this code is, it isn't in the onnx org - it's over in the Microsoft Org repositories [located here](https://github.com/microsoft/onnxconverter-common/tree/master/onnxconverter_common)

PR #283, opened 2019-04-16
Include onnxconverter-common tests, add a unit test on DataTypes

xadupre:


xadupre:
It fails for python 2.7. Do we need to support it?

vinitra:
Yes, I think we should continue to support Python 2.7. 

Looks like it's failing on a list.copy() since it doesn't exist in py2. I think we should use the b = list(a) method of copying instead, with a comment?

vinitra:
@xadupre, thanks for adding the onnxconverter-common tests to the CI build! LGTM.

PR #291, opened 2019-04-30
Added apply_slice function to enable multiple versions of Slice

prabhat00155:


onnxutils/onnxconverter_common/onnx_ops.py
```diff
@@ -497,6 +497,35 @@ def apply_scaled_tanh(scope, input_name, output_name, container, operator_name=N
         # output = a * d
         apply_mul(scope, [aName, dName], output_name, container)
 
+def apply_slice(scope, input_name, output_names, container, starts, ends,
```
vinitra: Since slice only has one output, call it output_name for consistency.

prabhat00155: Good point, fixed.

onnxutils/onnxconverter_common/onnx_ops.py
```diff
@@ -497,6 +497,37 @@ def apply_scaled_tanh(scope, input_name, output_name, container, operator_name=N
...
+        inputs.append(starts_name)
+        inputs.append(ends_name)
+        if axes:
+            axes_name = scope.get_unique_variable_name('axes')
+            container.add_initializer(axes_name, onnx_proto.TensorProto.INT64,
+                                      [len(axes)], axes)
+            inputs.append(axes_name)
+        if steps:
+            if not axes:
+                inputs.append('')
```
vinitra: In this case do you still need to add an initializer for axes?

prabhat00155: No, we are passing empty string as a placeholder for axes so that we can pass steps.

PR #313, opened 2019-06-10
fix import for scikit-learn >= 0.21

xadupre:


vinitra:
We need to remove the axis attribute from the Imputer call in CoreML's test_imputer. See line 15 of tests\coreml\test_cml_ImputerConverter.py.

```
    def test_imputer(self):
        model = Imputer(missing_values='NaN', strategy='mean', axis=0)
```

vinitra:
I think we'll have to remove NaN from the input testing data as well -- the new version of Imputer doesn't accept np.NaN as input

PR #329, opened 2019-08-13
feat: add support for lightgbm.Booster

hongzmsft:
For users who need to train a LightGBM model on huge datasets that do
not fit the memory, they often choose to train through the Booster API,
which supports two_round loading of huge datasets, or the command line,
which exports a model.txt file that can only be reconstructed to a
Booster. Currently the ONNX converter does not support either case.

Most of the converter work is about processing information from the
dumped model dictionary. Other information can also be inferred from
that information as well. Here we wrap the Booster information to
facilitate the conversion process.

### DIFF
--- a/onnxmltools/convert/lightgbm/_parse.py
+++ b/onnxmltools/convert/lightgbm/_parse.py
@@ -3,9 +3,11 @@
 # Licensed under the MIT License. See License.txt in the project root for
 # license information.
 # --------------------------------------------------------------------------
+import numpy
 
 from ..common._container import LightGbmModelContainer
 from ..common._topology import *
+from ..common.data_types import FloatTensorType
 
 from lightgbm import LGBMClassifier, LGBMRegressor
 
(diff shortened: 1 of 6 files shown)


hongzmsft:
There are a few issues that I need help:

1. I am not familiar with the test utility functions. Right now, I am not sure whether the current dump_data_and_model parameters are valid or not, especially regarding dump file names.

2. The current multiclass ONNX model has an issue with the dimension of the input to the ZipMap node. I could use some help in debugging that issue.

onnxmltools/convert/lightgbm/_parse.py
```diff
@@ -90,3 +123,7 @@ def parse_lightgbm(model, initial_types=None, target_opset=None,
         raw_model_container.add_output(variable)
 
     return topology
+
+
+def default_initial_types(model):
+    return [('features', FloatTensorType(shape=(1, model.n_features_)))]
```
xadupre: It should be return [('features', FloatTensorType(shape=(**None**, model.n_features_)))] as the number of rows is usually unknown. We recently introduced that change in onnxconverter-common.

onnxmltools/convert/lightgbm/convert.py
```diff
@@ -35,9 +40,10 @@ def convert(model, name=None, initial_types=None, doc_string='', target_opset=No
     :param custom_shape_calculators: a dictionary for specifying the user customized shape calculator
     :return: An ONNX model (type: ModelProto) which is equivalent to the input lightgbm model
     '''
+    if isinstance(model, lightgbm.Booster):
+        model = WrappedBooster(model)
     if initial_types is None:
-        raise ValueError('Initial types are required. See usage of convert(...) in \
-                           onnxmltools.convert.lightgbm.convert for details')
+        initial_types = default_initial_types(model)
```
xadupre: The documentation of the converter should reflect this change of behaviour.

onnxmltools/convert/lightgbm/_parse.py
hongzmsft: I have changed the code, but it seems the test does not handle it very well. Is it expected?

onnxmltools/convert/lightgbm/convert.py
hongzmsft: Docstring updated

onnxmltools/convert/lightgbm/_parse.py
hongzmsft: Well, I have retracted the automatic input inference part. It might need changes in code outside the scope of this PR. It would be better to make a separate PR for that.

onnxmltools/convert/lightgbm/_parse.py
```diff
@@ -90,3 +123,7 @@ def parse_lightgbm(model, initial_types=None, target_opset=None,
         raw_model_container.add_output(variable)
 
     return topology
+
+
+def default_initial_types(model):
```
xadupre: This function does not seem to be used anymore.

tests/lightgbm/test_LightGbmTreeEnsembleConverters.py
```diff
@@ -33,6 +35,33 @@ def test_lightgbm_regressor2(self):
...
+        X = numpy.array(X, dtype=numpy.float32)
+        y = [0, 1, 1.1]
+        data = lightgbm.Dataset(X, label=y)
+        model = lightgbm.train({'boosting_type': 'gbdt', 'objective': 'regression',
+                                'n_estimators': 3, 'min_child_samples': 1, 'max_depth': 1},
+                               data)
+        model_onnx, prefix = convert_model(model, 'tree-based binary classifier',
+                                           [('input', FloatTensorType([1, 2]))])
+        dump_data_and_model(X, model, model_onnx,
+                            basename=prefix + "Bin" + model.__class__.__name__)
```
xadupre: You can write basename=prefix + "BoosterBin" + model.__class__.__name__. The name is not really important except when the test fails. This name is used to dump everything on disk and investigate later.

onnxmltools/convert/lightgbm/_parse.py
hongzmsft: Removed

tests/lightgbm/test_LightGbmTreeEnsembleConverters.py
hongzmsft: Renamed

PR #342, opened 2019-10-09
Update sparkml unit test for onnxruntime 0.5+, shape[0]: 1 --> None

xadupre:


wenbingl:
The SparkML converter unit test has been disable from CI build now. And the default batch will be better to be 'N', instead of None.

PR #356, opened 2019-12-06
Fixing sklearn imputer discrepancy in v0.22

vinitra:
Imputer moved from preprocessing to impute sub-library in sklearn.


vinitra:
Tests here are not passing because of coremltools dependency. See https://github.com/onnx/sklearn-onnx/pull/337 for hotfix.

PR #357, opened 2019-12-09
Fix unit test with scikit-learn 0.22

xadupre:


requirements.txt
```diff
@@ -1,7 +1,7 @@
...
+keras2onnx
 numpy
-protobuf
 onnx
-skl2onnx
-keras2onnx
 onnxconverter-common>=1.5.0
+protobuf
+six
+skl2onnx
```
vinitra: thanks for alphabetizing!

tests/coreml/test_cml_ImputerConverter.py
```diff
@@ -1,20 +1,31 @@
...
+            # coremltools 3.1 does not work with scikit-learn 0.22
+            setattr(sklearn.preprocessing, 'Imputer', Imputer)
+        
+        try:
+            model = Imputer(missing_values='NaN', strategy='mean', axis=0)
+        except TypeError:
+            model = Imputer(missing_values='NaN', strategy='mean')
         data = [[1, 2], [np.nan, 3], [7, 6]]
         model.fit(data)
+        import coremltools  # noqa
```
vinitra: Why not put import coremltools after the sklearn import above?

xadupre: Just investigating.

xadupre: import coremltools fails if install with scikit-learn 0.22 so I moved it further. The test does not change but it fails later giving me more insights.

tests/coreml/test_cml_OneHotEncoderConverter.py
```diff
@@ -5,6 +5,14 @@
 import unittest
 import warnings
 import numpy
+try:
```
wenbingl: Can this code snippet be refactored as "Imputer = onnxmltools.sklearn.import_Imputer" to avoid a lot of duplication.

xadupre: No, I need to have coremltools Believe that Imputer is at a place it expects to find it.

PR #359, opened 2019-12-13
fixing example/tutorial bug

ksaur:
As written, the example gives an error:
```
[ONNXRuntimeError] : 2 : INVALID_ARGUMENT : Got invalid dimensions for input: float_input for the following indices
 index: 0 Got: 38 Expected: 1
 Please fix either the inputs or the model.
```

I'm submitting one possible fix.

Also, tutorial.rst was missing the `label_name` variable and numpy import (for float32) so I copied them as written in index.rst


docs/index.rst
```diff
@@ -55,7 +55,7 @@ to automatically check every converter with
     from onnxmltools import convert_sklearn
     from onnxmltools.utils import save_model
     from onnxconverter_common.data_types import FloatTensorType
-    initial_type = [('float_input', FloatTensorType([1, 4]))]
```
ksaur: `1` will not work here, you need to supply the batch size, but hardcoding it at 38 (from the iris dataset) may break later if the dataset changes, so `-1` supplies a the batch dimension.

Issue #371, opened 2020-02-26
XGBRegressor converter is maybe broken


victornoel:
It seems there are problems with the `XGBRegressor` converter.

I've opened an issue at sklearn-onnx but they redirected me to here: https://github.com/onnx/sklearn-onnx/issues/321

Here is a repro: https://gist.github.com/victornoel/06d6231f6276719ddba53cb381dfd468

I casted an int column to str because in my original dataset, we have categories made of numbers so I thought this could be related…

As we can see, the differences are huge: the predictions should be between 0 and 1 and the differences can be as big as 1 in my tests!
```
>>> import test
>>> test.test()
[0.9019426  0.91082716 0.91868186 0.94926846 1.0898147 ]
min(Y)-max(Y): 0 1
```

victornoel:
For the record, I tested latest git version of sklearn-onnx, onnxconverter-common and onnxmltools and the problem is still there…

xadupre:
I was able to replicate. I'll have a look tomorrow.

xadupre:
One issue comes from OneHotEncoder. By default, scikit-learn is using sparse but sparse are not supported yet in Onnxruntime even though they are defined in ONNX. Because Onnxruntime is using a dense representation, missing values are different in scikit-learn (nan) and dense matrices in onnxruntime (0). I suggest using OneHotEncoder(sparse=False) for the time being until sparse are supported in onnxruntime and fully supported in ONNX.

victornoel:
@xadupre thank you for the investigation, I have some questions:
- using sparse does indeed fix the problem for me
- would it make sense for sklearn-onnx to fail for `OneHotEncoder` that were built with `sparse=True` maybe to avoid any problem until this is supported?
- what are the drawbacks of sparse? It feels to me like training a model takes much more time with `sparse=False`…
- is there an issue to track supporting sparse in onnxruntime?
- you say "one issue", are there more ? :smile:

xadupre:
I'm hesitating. sklearn-onnx is being tested with Onnxruntime but it does not mean only Onnxruntime should run ONNX graphs, so failing for OneHotEncoder is not my favourite option. For the OneHotEncoder, there is usually no drawback as it saves much memory space and it is faster as many features are null and are not part of the computation. So, in this case, sparse is definitly better. There is an issue on onnx (https://github.com/onnx/onnx/issues/2008), none in onnxruntime. Feel free to add one. I don't have a better option right now.

victornoel:
[...]

Issue #368, opened 2020-01-31
Support for XGBoost 1.0.0


victornoel:
This is maybe a bit early but XGBoost 1.0.0rc1 (see https://github.com/dmlc/xgboost/issues/5253) was just released and it seems that it breaks enough things so that it doesn't work when using `onnxmltools.convert.xgboost.operator_converters.XGBoost.convert_xgboost`.

This is what happens when using `convert_sklearn` on a pipeline containing an `XGBoostRegressor`:
```
 File "/home/victor/.cache/pypoetry/virtualenvs/env-51Be8MFl-py3.7/lib/python3.7/site-packages/skl2onnx/convert.py", line 138, in convert_sklearn
    dtype=dtype, options=options)
  File "/home/victor/.cache/pypoetry/virtualenvs/env-51Be8MFl-py3.7/lib/python3.7/site-packages/skl2onnx/common/_topology.py", line 963, in convert_topology
    conv(scope, operator, container)
  File "/home/victor/.cache/pypoetry/virtualenvs/env-51Be8MFl-py3.7/lib/python3.7/site-packages/onnxmltools/convert/xgboost/operator_converters/XGBoost.py", line 279, in convert_xgboost
    cls.convert(scope, operator, container)
  File "/home/victor/.cache/pypoetry/virtualenvs/env-51Be8MFl-py3.7/lib/python3.7/site-packages/onnxmltools/convert/xgboost/operator_converters/XGBoost.py", line 189, in convert
    name=scope.get_unique_operator_name('TreeEnsembleRegressor'), **attr_pairs)
  File "/home/victor/.cache/pypoetry/virtualenvs/env-51Be8MFl-py3.7/lib/python3.7/site-packages/skl2onnx/common/_container.py", line 383, in add_node
    "".format(op_type, name)) from e
ValueError: Unable to create node 'TreeEnsembleRegressor' with name='TreeEnsembleRegressor'.
```

For the record it works with XGBoost 0.90 and `convert_xgboost` is setup with `skl2onnx.update_registered_converter`.

trivialfis:
I can help taking a look if there's a reproducible script.  Currently it's still RC so we can fix it.

victornoel:
@trivialfis I think https://github.com/onnx/sklearn-onnx/blob/master/tests_onnxmltools/test_xgboost_converters.py should be perfect for this

trivialfis:
Alright.  Will take a look later tonight.  Thanks for the link!

[...]

PR #373, opened 2020-03-04
Fix xgboost converter

xadupre:
Investigate issue #372, #371, #368. It looks like missing value are not supported the same way in xgboost and in onnxruntime.


requirements.txt
```diff
@@ -1,7 +1,7 @@
 keras2onnx
 numpy
 onnx
-onnxconverter-common>=1.6.5, <1.7.0
```
wenbingl: Don't change that, this is mandatory, according to LT team's comments.

xadupre: Fixed. I don't remember the explicit Reason though.

PR #381, opened 2020-04-02
Fix for onnx 1.7 release

wenbingl:


onnxmltools/proto/__init__.py
```diff
@@ -51,4 +55,13 @@ def _make_tensor_fixed(name, data_type, dims, vals, raw=False):
...
 
 def get_opset_number_from_onnx():
-    return onnx.defs.onnx_opset_version()
+    # since the method was widely used among while it is buggy to get the opset number...
+    # ... blindly, so change it to be safer without the name change.
+
+    default_max_opset = 11
+    try:
+        from onnxconverter_common.topology import DEFAULT_OPSET_NUMBER
+        default_max_opset = DEFAULT_OPSET_NUMBER
```
jiafatom, line 64: Shall we do `min(default_max_opset, DEFAULT_OPSET_NUMBER)` ? Otherwise if onnxconverter_common has larger opset, then it makes no sense.

PR #391, opened 2020-05-15
CatBoost converter

monkey0head:
Hi! I want to add [CatBoost](https://catboost.ai/) models conversion functionality. 

CatBoost has its own [converter](https://catboost.ai/docs/concepts/apply-onnx-ml.html) to ONNX-ML and I added an interface to convert CatBoost models with onnxmltools as it would be more convenient for users to be able to convert CatBoost models in the same way as models created with other popular ml toolkits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,7 @@ ONNXMLTools enables you to convert models from different machine learning toolki
 * libsvm
 * XGBoost
 * H2O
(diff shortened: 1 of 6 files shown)


monkey0head:
I will open new pr to resolve CLA issue

PR #392, opened 2020-05-15
CatBoost converter

monkey0head:
Hi! I want to add [CatBoost](https://catboost.ai/) models conversion functionality. 

CatBoost has its own [converter](https://catboost.ai/docs/concepts/apply-onnx-ml.html) to ONNX-ML and I added an interface to convert CatBoost models with onnxmltools as it would be more convenient for users to be able to convert CatBoost models in the same way as models created with other popular ml toolkits.

### DIFF
--- a/onnxmltools/convert/main.py
+++ b/onnxmltools/convert/main.py
@@ -7,7 +7,6 @@
 import onnx
 from .common import utils
 import warnings
-import importlib
 
 
 def convert_coreml(model, name=None, initial_types=None, doc_string='', target_opset=None,
@@ -177,9 +176,8 @@ def convert_tensorflow(frozen_graph_def,
                        target_opset=None,
                        channel_first_inputs=None,
                        debug_mode=False, custom_op_conversions=None):
-    try:
-        importlib.import_module('tf2onnx')
-    except (ImportError, ModuleNotFoundError) as e:
+    import pkgutil
+    if not pkgutil.find_loader('tf2onnx'):
         raise RuntimeError('tf2onnx is not installed, please install it before calling this function.')
 
     return _convert_tf_wrapper(frozen_graph_def, name, input_names, output_names, doc_string,
--- a/onnxmltools/convert/main.py
+++ b/onnxmltools/convert/main.py
@@ -44,27 +44,15 @@ def convert_libsvm(model, name=None, initial_types=None, doc_string='', target_o
                    custom_conversion_functions, custom_shape_calculators)
 
 
-def convert_catboost(model, name=None, initial_types=None, doc_string='', target_opset=None,
-                   targeted_onnx=onnx.__version__, custom_conversion_functions=None, custom_shape_calculators=None):
+def convert_catboost(model, name=None, initial_types=None, doc_string='', target_opset=None):
     try:
         from catboost.utils import convert_to_onnx_object
     except ImportError:
-        raise RuntimeError('CatBoost is not installed or need to be updated. '
+        raise RuntimeError('CatBoost is not installed or needs to be updated. '
                            'Please install/upgrade CatBoost to use this feature.')
 
-    if custom_conversion_functions:
-        warnings.warn('custom_conversion_functions is not supported. Please set it to None.')
-    if custom_shape_calculators:
-        warnings.warn('custom_shape_calculators is not supported. Please set it to None.')
-
-    export_parameters = {
-        'onnx_domain': 'ai.catboost',
-        'onnx_model_version': 0,
-        'onnx_doc_string': doc_string,
-        'onnx_graph_name': name
-    }
-
-    return convert_to_onnx_object(model, export_parameters=export_parameters)
+    return convert_to_onnx_object(model, export_parameters={'onnx_doc_string': doc_string, 'onnx_graph_name': name},
+                                  initial_types=initial_types, target_opset=target_opset)
 
 
 def convert_lightgbm(model, name=None, initial_types=None, doc_string='', target_opset=None,
--- a/onnxmltools/convert/main.py
+++ b/onnxmltools/convert/main.py
@@ -44,6 +44,29 @@ def convert_libsvm(model, name=None, initial_types=None, doc_string='', target_o
                    custom_conversion_functions, custom_shape_calculators)
 
 
+def convert_catboost(model, name=None, initial_types=None, doc_string='', target_opset=None,
+                   targeted_onnx=onnx.__version__, custom_conversion_functions=None, custom_shape_calculators=None):
+    try:
+        from catboost.utils import convert_to_onnx_object
+    except ImportError:
+        raise RuntimeError('CatBoost is not installed or need to be updated. '
+                           'Please install/upgrade CatBoost to use this feature.')
+
+    if custom_conversion_functions:
+        warnings.warn('custom_conversion_functions is not supported. Please set it to None.')
+    if custom_shape_calculators:
+        warnings.warn('custom_shape_calculators is not supported. Please set it to None.')
+
+    export_parameters = {
+        'onnx_domain': 'ai.catboost',
+        'onnx_model_version': 0,
+        'onnx_doc_string': doc_string,
+        'onnx_graph_name': name
+    }
+
+    return convert_to_onnx_object(model, export_parameters=export_parameters)
+
+
 def convert_lightgbm(model, name=None, initial_types=None, doc_string='', target_opset=None,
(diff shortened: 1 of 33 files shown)


vinitra:
Thanks for the contribution, @monkey0head!

tests/catboost/test_CatBoost_converter.py
```diff
@@ -0,0 +1,53 @@
...
+        catboost_model = catboost.CatBoostClassifier(task_type='CPU', loss_function='CrossEntropy',
+                                                     n_estimators=10, verbose=0)
+
+        catboost_model.fit(X.astype(numpy.float32), y)
+
+        catboost_onnx = convert_catboost(catboost_model, name='CatBoostBinClassification',
+                                         doc_string='test binary classification')
+        self.assertTrue(catboost_onnx is not None)
+        # onnx runtime returns zeros as class labels
+        # dump_data_and_model(X.astype(numpy.float32), catboost_model, catboost_onnx, basename="CatBoostBinClass")
```
vinitra, line 36: Should this line be uncommented?

onnxmltools/convert/main.py
```diff
@@ -44,6 +44,29 @@ def convert_libsvm(model, name=None, initial_types=None, doc_string='', target_o
...
+def convert_catboost(model, name=None, initial_types=None, doc_string='', target_opset=None,
+                   targeted_onnx=onnx.__version__, custom_conversion_functions=None, custom_shape_calculators=None):
+    try:
+        from catboost.utils import convert_to_onnx_object
+    except ImportError:
+        raise RuntimeError('CatBoost is not installed or need to be updated. '
+                           'Please install/upgrade CatBoost to use this feature.')
+
+    if custom_conversion_functions:
+        warnings.warn('custom_conversion_functions is not supported. Please set it to None.')
```
vinitra, line 56: Why include these converter arguments if they are not supported? It might be better to remove the arguments entirely. In the code above for the keras converter, these arguments were deprecated, which is why the warning messages were necessary.

onnxmltools/convert/main.py
```diff
@@ -44,6 +44,29 @@ def convert_libsvm(model, name=None, initial_types=None, doc_string='', target_o
                    custom_conversion_functions, custom_shape_calculators)
 
 
+def convert_catboost(model, name=None, initial_types=None, doc_string='', target_opset=None,
+                   targeted_onnx=onnx.__version__, custom_conversion_functions=None, custom_shape_calculators=None):
+    try:
+        from catboost.utils import convert_to_onnx_object
+    except ImportError:
+        raise RuntimeError('CatBoost is not installed or need to be updated. '
```
vinitra, line 52: nit: "needs to be updated."

onnxmltools/convert/main.py
monkey0head: I thought all convertors have pretty the same interface and thus added the args). I have discussed the matter with the member of CatBoost team. I will create a pr to change CatBoost converter interface to pass those args to the CatBoost's side. CatBoost team may implement the functionality in the future. I will update my pr when the change is released if it is ok.

tests/catboost/test_CatBoost_converter.py
monkey0head: Well, this part has a problem :( 
The comparison works properly with probabilities, not with labels. A converted model returns only zeros as labels.  I consulted the Catboost team and they consider it as onnxruntime bug.

onnxmltools/convert/main.py
xadupre: The signature in onnxmltools is not always the same. Only in sklearn-onnx. So I would either remove the parameter either raise an exception if the parameter is not None.

onnxmltools/convert/main.py
```diff
@@ -44,6 +44,29 @@ def convert_libsvm(model, name=None, initial_types=None, doc_string='', target_o
...
+                           'Please install/upgrade CatBoost to use this feature.')
+
+    if custom_conversion_functions:
+        warnings.warn('custom_conversion_functions is not supported. Please set it to None.')
+    if custom_shape_calculators:
+        warnings.warn('custom_shape_calculators is not supported. Please set it to None.')
+
+    export_parameters = {
+        'onnx_domain': 'ai.catboost',
```
xadupre, line 61: Why not using existing domains?

tests/catboost/test_CatBoost_converter.py
xadupre: This must be fixed and it is probably an error somewhere in the onnx graph.

monkey0head: Thank you for the information, I reported your reply to the Catboost team members and I will update my pr after they fix the bug

onnxmltools/convert/main.py
```diff
@@ -44,6 +44,29 @@ def convert_libsvm(model, name=None, initial_types=None, doc_string='', target_o
                    custom_conversion_functions, custom_shape_calculators)
 
 
+def convert_catboost(model, name=None, initial_types=None, doc_string='', target_opset=None,
+                   targeted_onnx=onnx.__version__, custom_conversion_functions=None, custom_shape_calculators=None):
```
wenbingl, line 48: The other converters keeps  arguments like "targeted_onnx=onnx.__version__, custom_conversion_functions=None, custom_shape_calculators=None" for the backward compatibility, if there is a brand new one, these arguments could be dropped.

onnxmltools/utils/tests_helper.py
```diff
@@ -212,6 +212,9 @@ def convert_model(model, name, input_types):
             model, prefix = convert_lightgbm(model, name, input_types), "LightGbm"
         else:
             raise RuntimeError("Unable to convert model of type '{0}'.".format(type(model)))
+    elif model.__class__.__name__.startswith("Cat"):
```
wenbingl, line 215: Any better fingerprint to identify the original model?

onnxmltools/convert/main.py
```diff
@@ -44,6 +44,29 @@ def convert_libsvm(model, name=None, initial_types=None, doc_string='', target_o
...
+        warnings.warn('custom_shape_calculators is not supported. Please set it to None.')
+
+    export_parameters = {
+        'onnx_domain': 'ai.catboost',
+        'onnx_model_version': 0,
+        'onnx_doc_string': doc_string,
+        'onnx_graph_name': name
+    }
+
+    return convert_to_onnx_object(model, export_parameters=export_parameters)
```
wenbingl, line 67: Need take care of the target_opset argument, which specify what's the opset version will be used in the generated ONNX model.
If you plan to only support one target_opset currently, you need check target_opset and report an issue if the user target_opset is not as same as the one that is support in Catboost.

tests/catboost/test_CatBoost_converter.py
monkey0head: It works now with the new onnxruntime version

onnxmltools/convert/main.py
monkey0head: Thanks, I removed arguments that are not supported

onnxmltools/convert/main.py
monkey0head: removed

onnxmltools/convert/main.py
monkey0head: Thanks, now I pass the target_opset to Catboost and check it there

onnxmltools/convert/main.py
monkey0head: Thanks, did so

monkey0head:
There was a problem with binary classification: https://github.com/onnx/onnxmltools/pull/392#discussion_r426199488

The converter for binary classification works well now with the new version of onnxruntime (tested locally and tests for rt130 are passed here) but fails for the older versions.

monkey0head:
Hi! Could you please have a look at the updated pull request?

Regarding the [binary classification issue](https://github.com/onnx/onnxmltools/pull/392#discussion_r426199488): The converter for binary classification works well with the new version of onnxruntime. The checks fail on the older versions.

wenbingl:
I saw some failure in CI pipeline which blocks your PR merging. Can you fix them before merge the PR?

monkey0head:
> I saw some failure in CI pipeline which blocks your PR merging. Can you fix them before merge the PR?

@wenbingl, sure. Fixed!

nikitxskv:
@vinitra @wenbingl @xadupre 
Hi!
Can you tell us (CatBoost team) what should be done to be added to the list of supported tools [here](https://onnx.ai/supported-tools.html)?

PR #397, opened 2020-05-20
Fixes #396, xgboost converter for xgboost >= 1.0.2

### LINKED ISSUE #396: onnxmltools fails with the latest version of XGBoost

xadupre:


onnxmltools/convert/xgboost/common.py
```diff
@@ -12,5 +12,9 @@ def get_xgb_params(xgb_node):
     else:
         # XGBoost < 0.7
         params = xgb_node.__dict__
-        
+
+    if ('n_estimators' not in params and
```
jiafatom, line 16: no need to have ()

xadupre: I need to check because I'm not sure which versions of python support this. I use that because that's what's done in scikit-learn.

xadupre: It does not work without () for me.

wenbingl: Is it due to parenthesis has to enclose the (not in)? not the full statement?

xadupre: The parenthesis are used to write the test over two lines, otherwise, I would have to use symbol \. Apparently, if ( … \n ...): seems to be the preferred one.

PR #416, opened 2020-08-20
Install cerberus before ort-nightly

jiafatom:
nightly build fails due to a recent onnxruntime change that add `cerberus` into requirement.txt. We need install the dependency in a separate step before installing ort-nightly. This PR does it.


.azure-pipelines/linux-CI-nightly.yml
```diff
@@ -42,6 +42,7 @@ jobs:
       python -m pip install git+https://github.com/onnx/keras-onnx
       python -m pip install -r requirements.txt
       python -m pip install -r requirements-dev.txt
+      python -m pip install cerberus
```
xadupre, line 45: why not adding cerberus to requirements-dev.txt?

xadupre (changes requested):
Why not using requirements-dev.txt?

PR #423, opened 2020-09-27
Update XGBoost.py

clhne:
When the type of feature_id is float there throw an error:
"ValueError: invalid literal for int() with base 10:"
Because you get a ValueError if you pass a string representation of a float into int, or a string representation of anything but an integer (including empty string). If you do want to pass a string representation of a float to an int, as it points out above, you can convert to a float first, then to an integer.
Ref: https://stackoverflow.com/questions/1841565/valueerror-invalid-literal-for-int-with-base-10


clhne:
@xadupre @wenbingl @sdpython @prabhat00155 @msteijaert 
Please review, thanks.

xadupre (approved):

xadupre (approved):

Issue #376, opened 2020-03-18
Can't convert Core ML model to Onnx


aof-github:
Hello,

I'm trying to convert a trained Core ML model to Onnx in order to convert it to TensorFlow Lite.
The problems is that I get errors. I've tried with different versions of python, onnxmltools, winmltools and it doesn't seems to work. I also tried docker image of onnx ecosystem with same result. Can any one help me with it? Thanks in advance.

** Script **
import coremltools
import onnxmltools

# Update the input name and path for your CoreML model
input_coreml_model = '../ClasificadorGolpesIzquierda.mlmodel'

# Change this path to the output name and path for the ONNX model
output_onnx_model = '../ClasificadorGolpesIzquierda.onnx'

# Load your CoreML model
coreml_model = coremltools.utils.load_spec(input_coreml_model)

# Convert the CoreML model into ONNX
onnx_model = onnxmltools.convert_coreml(coreml_model)

# Save as protobuf
onnxmltools.utils.save_model(onnx_model, output_onnx_model)

**Error Messages**
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-11-94a6dc527869> in <module>
      3 
      4 # Convert the CoreML model into ONNX
----> 5 onnx_model = onnxmltools.convert_coreml(coreml_model)
      6 
      7 # Save as protobuf

/usr/local/lib/python3.6/dist-packages/onnxmltools/convert/main.py in convert_coreml(model, name, initial_types, doc_string, target_opset, targeted_onnx, custom_conversion_functions, custom_shape_calculators)
     16     from .coreml.convert import convert
     17     return convert(model, name, initial_types, doc_string, target_opset, targeted_onnx,
---> 18                    custom_conversion_functions, custom_shape_calculators)
     19 
     20 

/usr/local/lib/python3.6/dist-packages/onnxmltools/convert/coreml/convert.py in convert(model, name, initial_types, doc_string, target_opset, targeted_onnx, custom_conversion_functions, custom_shape_calculators)
     58     target_opset = target_opset if target_opset else get_opset_number_from_onnx()
     59     # Parse CoreML model as our internal data structure (i.e., Topology)
---> 60     topology = parse_coreml(spec, initial_types, target_opset, custom_conversion_functions, custom_shape_calculators)
     61 
     62     # Parse CoreML description, author, and license. Those information will be attached to the final ONNX model.

[...]

Issue #331, opened 2019-08-14
convert coreml to onnx


sharepointer:
onnx_model = onnxmltools.convert_coreml(coreml_model)
  File "/usr/local/lib/python3.6/dist-packages/onnxmltools/convert/main.py", line 18, in convert_coreml
    custom_conversion_functions, custom_shape_calculators)
  File "/usr/local/lib/python3.6/dist-packages/onnxmltools/convert/coreml/convert.py", line 80, in convert
    onnx_model = convert_topology(topology, name, doc_string, target_opset, targeted_onnx)
  File "/usr/local/lib/python3.6/dist-packages/onnxconverter_common/topology.py", line 729, in convert_topology
    registration.get_converter(operator.type)(scope, operator, container)
  File "/usr/local/lib/python3.6/dist-packages/onnxmltools/convert/coreml/operator_converters/neural_network/Activation.py", line 27, in convert_activation
    apply_prelu(scope, inputs, outputs, container, operator_name=attrs['name'], slope=[params.PReLU.alpha])
  File "/usr/local/lib/python3.6/dist-packages/onnxconverter_common/onnx_ops.py", line 440, in apply_prelu
    s_shape = slope.shape
AttributeError: 'list' object has no attribute 'shape'

sharepointer:
error on apply_prelu???

jiafatom:
Can you provide the model for us to debug?

KeunwooPark:
I had same issue. I was trying to convert [Openpose](https://github.com/CMU-Perceptual-Computing-Lab/openpose) to an ONNX model. I used their body_25 models. The below is my code.

```python
# Convert Caffe model to CoreML
coreml_model = coremltools.converters.caffe.convert((args.caffemodel, args.protofile))
# Convert the Core ML model into ONNX
onnx_model = onnxmltools.convert_coreml(coreml_model)
# Save as protobuf
onnxmltools.utils.save_model(onnx_model, args.output)
```
And here is the error I got.

```shell
Traceback (most recent call last):
  File "caffe2onnx.py", line 27, in <module>
    main(args)
  File "caffe2onnx.py", line 21, in main
    onnx_model = onnxmltools.convert_coreml(coreml_model)
  File "[home]/.virtualenvs/caffe2pytorch/lib/python3.5/site-packages/onnxmltools/convert/main.py", line 18, in convert_coreml
    custom_conversion_functions, custom_shape_calculators)
[...]

PR #425, opened 2020-10-14
Fix PReLU conversion from CoreML

dudalev:
This fixes issues #331 & #376, but there are potentially more issues in other activation functions


Laubeee:
This does *not* fix #376 . It does fix the first error I encountered, but there just pops up a next one instead. Check the latest post(s) in the issue

wenbingl (approved):

PR #427, opened 2020-10-28
Fix for API changes in Hummingbird 0.1.0

interesaaat:
In the next release of Hummingbird we will deprecate the ability to pass input types to the converter. This PR makes sure that the the lightGBM converter will work with the new Hummingbird API.


interesaaat:
I pinned Hummingbird to version 0.0.6 (the current version) instead of changing the API.

interesaaat:
@wenbingl we are going to push the new Hummingbird release today. In case your nightly breaks, this PR will fix it.

wenbingl (approved):

PR #429, opened 2020-11-06
Fix the converter when parameter best_ntree_limit is used

xadupre:
Fix for issue https://github.com/microsoft/onnxruntime/issues/5450.


wenbingl (approved):

nsiddharth:
Wonder if the latest version has this fix? if not, is there some estimate on when this might be merged and become available?

wenbingl (approved):

PR #442, opened 2021-01-20
Fix #400, support multi:softmax objective

### LINKED ISSUE #400: ONNX does not support "multi:softmax" of XgBoost
Hi, 
first I wanted to thank you for the nice library. 

I am trying to convert XgBoost mult-nomial/multi-class logistic regression to ONNX. Checking the source code I have noticed that the file "_parse.py" does not support "multi:softmax", but only "multi:softprob". Indeed, the predictions are all logits and not the actual classes. The predictions I get with "multi:softmax" are exactly the same the ones I get by using "multi:softprob". I think the "multi:softmax" gets overridden automatically by ONNX. 

Thanks, 
Paolo

On Ubuntu 18.04, I am using: 
onnx==1.7.0
xgboost==1.1.0

xadupre:
Signed-off-by: xavier dupré <xavier.dupre@gmail.com>


.azure-pipelines/win32-conda-CI.yml
```diff
@@ -1,113 +1,113 @@
-# Python package
```
wenbingl, line 1: What's the change in the file? all lines look updated

xadupre: Windows probably changed the end of line. I need to restore it.

xadupre: Fixed.

wenbingl (approved):

Issue #451, opened 2021-03-11
enable convert_lightgbm to output tensor type


huzq2016:
**Description**
When I used convert_lightgbm, I got a neural network:
<img width="198" alt="lightgbm" src="https://user-images.githubusercontent.com/23181889/110765870-9237d980-828f-11eb-966c-79a108639cb7.png">
which includes a probabilities zipmap type that is not supported by onnxruntime server so far (see: https://github.com/microsoft/onnxruntime/issues/2385)

As comparison, when I used convert_xgboost, I got a neural network:
<img width="185" alt="xgboost" src="https://user-images.githubusercontent.com/23181889/110766285-f5297080-828f-11eb-8df1-b3de5d36e081.png">
which includes a probabilities tensor that is supported by onnxruntime server.

Very probably the zipmap type in convert_lightgbm is caused by https://github.com/onnx/onnxmltools/blob/master/onnxmltools/convert/lightgbm/operator_converters/LightGbm.py#L454.
In convert_xgboost, it is https://github.com/onnx/onnxmltools/blob/master/onnxmltools/convert/xgboost/operator_converters/XGBoost.py#L291, which seemingly does not have zipmap.

**Describe the solution you'd like**
Conversion of lightgbm into onnx models can output the tensor type of probabilities.

wenbingl:
@xadupre , can you help with that?

xadupre:
It is possible to remove zipmap operator by using options. The example shows how to use sklearn-onnx for lightgbm [Convert a pipeline with a LightGBM model](http://onnx.ai/sklearn-onnx/auto_tutorial/plot_gexternal_lightgbm.html). The second one shows how to remove zipmap node [One model, many possible conversions with options](http://onnx.ai/sklearn-onnx/auto_tutorial/plot_dbegin_options.html#option-zipmap).

huzq2016:
thanks @xadupre. 
The example is only for sklearn-onnx. Wonder if a similar solution can be implemented for convert_lightgbm of onnxmltools?

xadupre:
It could be but I could probably choose a more simple way, just adding a parameter `converrt_lightgbm(..., zipmap=True)`.

huzq2016:
@xadupre, actually we have tried sklearn-onnx, but always failed due to some issue. So it is better if we can implemented here.

sdpython:
Ok I have two bugs to fix then. I'll start by adding a parameter zipmap to convert_lightgbm.

[...]

PR #452, opened 2021-03-12
Enable option zipmap for LGBM converter (fix issue #451)

xadupre:


wenbingl:
What happens on this PR? I saw the big changes on every files. How do you suppose I can review the change?

xadupre:
Sorry for that. I forgot to check the end of line. I wonder if there is a way to add a unit test checking that. I'll change that.

xadupre:
It looks better now.

wenbingl (approved):

PR #464, opened 2021-05-28
Don't append threshold twice

janjagusch:
When trying to convert a LightGBM model that uses categorical variables with `onnxmltools` and then loading it back with `onnxruntime`, I was getting a strange error message:

`ValueError: You passed in an iterable attribute but I cannot figure out its applicable type.`

This happened because the `attrs["node_values"]` property contained float and string values.

```
set([type(val) for val in attrs["node_values"]])
{<class 'str'>, <class 'float'>}
```

The reason for that is because if `tree_structure['threshold']` is a string, it is appended to `attrs['nodes_values']` twice - once as float and once as string.

I fixed it so that I only gets appended once and now it's working. 😄


janjagusch:
@wenbingl, if you have the time, could you maybe take a look at this, please?

wenbingl (approved):

wenbingl:
Thanks for the fixing!

PR #473, opened 2021-06-22
Remove outdated HB config setting

ksaur:
Thanks to the PR (https://github.com/microsoft/hummingbird/pull/523) from @xadupre , we should be able to remove the version pin for Hummingbird.


xadupre:
This PR probably requires #471 to be merged first to fix some issues with lightgbm.

ksaur:
In addition to needing that PR, we have some additional work to do in the LGBM code (fixing an API change).

ksaur:
Installing `torch==1.9.x` (most current) is giving a memory error in the az pipeline; i'm not sure what to do about that.

xadupre:
#471 was merged. It is worth updating the PR and try again.

ksaur:
As it is now, this PR just removes the outdated line, already commented out.  

Closing.

PR #463, opened 2021-05-11
Allow to add custom post transform functions that are not supported by the ONNX spec yet

janjagusch:
Closes #462 

I have already added support for `objective='poisson'`, but I believe this PR also allows to easily add support for `objective='tweedie'` and `objective='gamma'`.

Ultimately, it would be useful to have "EXPONENTIAL" as a supported value in `attrs['post_transform']`, such that we don't have to handle this special case here. However, that seems to involve quite some changes in `onxx`/`onnxruntime`. If anybody knows how to move these changes upstream, that would be very appreciated.

What this PR is still missing is tests. I have tested this locally, asserting that the predictions of the LightGMB model and the ONNX model are equivalent. I don't see an easy way to add this to your test suite, though. Some advice on this would be great.

PR #478, opened 2021-07-02
add node to container instead of changing output node of container

janjagusch:
Hi!

In #463 I added a bug (sorry about that!) by [changing that output node of the container](https://github.com/onnx/onnxmltools/pull/463/files#diff-6bd8fd0d594bbe670b1e60cd0f196ef22f7ad9d54a8b60122e4352438789a9c9R420). This is a really bad idea and can lead to strange side effects where this node will remain a container output, even after adding more nodes that use it as input.

Generally I believe it's much safer to use the `container.add_node` function instead, which is what I'm doing in this PR.


xadupre (approved):

janjagusch:
@xadupre thanks for approving! could you also merge it, please? 🙂

PR #488, opened 2021-08-09
Fix missing type in lgbm regressor

janjagusch:
Fixed a bug, where the ONNX model would take a different path through the tree compared to the LGBM regressor when `tree_structure["missing_type"] == 'None' and float(tree_structure['threshold']) < 0.0`.


janjagusch:
@xadupre, if you have time, could you please take a look at this? 🙂 

it's quite important, as the ONNX graph currently produces wrong predictions in some special cases.

xadupre (approved):

PR #491, opened 2021-08-18
Improves lightgbm conversion speed

### LINKED ISSUE #482: convert_lightgbm appears to be stuck in an infinite loop for a multiclass model
Python 3.8.8
onnxmltools 1.7.0 (last we checked, higher versions break the output data type we need--the was bug fixed and then reintroduced)

We have an example that has been running/stuck for 4.5 days.
I started another run yesterday with a debugger attached. When I pause the process, I see a deep stack with modify_tree_for_rule_in_set & it's recursive_call method:
image

The lightgbm model.txt file is 251MB (Attachment of the full file failed--I assume because it was too big, but I've attached a file containing just the 1st 10 trees. Please reach out for the model.txt if needed for repro).
[multiclass_lgbm_model_truncated_trees.txt](https://github.com/onnx/onnxmltools/files/6858215/multiclass_lgbm_model_truncated_trees.txt)

xadupre:
Conversion speed up for lightgbm model. For a big model (~40000 trees), the conversion goes from 3 hours to 10 minutes for lightgbm. Fixes #482.

### DIFF
--- a/.azure-pipelines/linux-conda-CI.yml
+++ b/.azure-pipelines/linux-conda-CI.yml
@@ -91,6 +91,7 @@ jobs:
     displayName: 'Install dependencies'
 
   - script: |
+      pip install flake8
       python -m flake8 ./onnxmltools
     displayName: 'run flake8 check'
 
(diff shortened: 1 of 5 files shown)


abhishek9804:
can anyone tell me why i am getting below error on python 3.8, during lightgbm model.
dump_booster_model
    ret = json.loads(string_buffer.value.decode('utf-8'), cls=Hook,
TypeError: json_loads_mock() got an unexpected keyword argument 'cls'.

PR #493, opened 2021-08-23
Add requirements.txt to MANIFEST.in

xhochy:
Thanks @xadupre for uploading the `sdist`. Sadly I cannot build from it as the `requirements.txt` file is not part of it and the `setup.py` requires it. The following PR should fix it.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,3 @@
 include onnxmltools/utils/*.js
 include onnxmltools/utils/*.css
+include requirements.txt


janjagusch:
@xadupre, would be great if you could release another tag after this is released. 🙂

xadupre (approved):

xadupre:
I published a new release https://github.com/onnx/onnxmltools/releases/tag/1.9.1, let me know if this one works for you.

PR #503, opened 2021-09-23
add support for quantile objective for LGBM models

muyajil:
This PR adds support to directly convert LGBM models that use the objective 'quantile'.
There is no real difference in handling the output of a quantile model versus a regression model, since the quantile model is just a regression tuned to output a certain quantile, therefore it can be handled the same.


janjagusch:
thanks for this PR!

i'm not a maintainer of this repo but i implemented a bit of stuff related to the LGBM objectives previously. i would suggest to add a test case [here](https://github.com/onnx/onnxmltools/blob/master/tests/lightgbm/test_objective_functions.py#L30-L34) to make sure that your changes are working as expected. 🙂

xadupre (approved):

Issue #499, opened 2021-09-18
Error converting deserialized xgboost Booster


ide-an:
Hello.

I am trying to convert existing xgboost model file (which is created by `xgboost.Booster.save_model`) to onnx. While doing that I am getting the following error:

`AttributeError: 'Booster' object has no attribute 'best_ntree_limit'`

my environment
 * OS: Windows 10 Home
 * xgboost==1.4.2
 * onnxmltools==1.9.1
 * onnx==1.10.1

reproduction code:
```python
from sklearn.datasets import load_iris
from sklearn.model_selection import train_test_split
from xgboost import DMatrix, Booster, train as train_xgb
from onnxconverter_common.data_types import FloatTensorType
from onnxmltools.convert import convert_xgboost

# Train
iris = load_iris()
X, y = iris.data, iris.target
X_train, X_test, y_train, y_test = train_test_split(X, y)
dtrain = DMatrix(X_train, label=y_train)
param = {'objective': 'multi:softmax', 'num_class': 3}
bst_original = train_xgb(param, dtrain, 10)
initial_type = [('float_input', FloatTensorType([None, 4]))]

# Converting original Booster is OK.
onx = convert_xgboost(bst_original, initial_types=initial_type)

# Save and load model
bst_original.save_model('model.json')
bst_loaded = Booster()
bst_loaded.load_model('model.json')

# !!! Converting loaded Booster fails !!!
onx_loaded = convert_xgboost(bst_loaded, initial_types=initial_type)
```

stack trace:
```
[01:32:25] WARNING: C:/Users/Administrator/workspace/xgboost-win64_release_1.4.0/src/learner.cc:1095: Starting in XGBoost 1.3.0, the default evaluation metric used with the objective 'multi:softmax' was changed from 'merror' to 'mlogloss'. Explicitly set eval_metric if you'd like to restore the old behavior.
Traceback (most recent call last):
  File "conv.py", line 25, in <module>
    onx_loaded = convert_xgboost(bst_loaded, initial_types=initial_type)
  File "C:\Users\macke\repos\onnx_sample\env\lib\site-packages\onnxmltools\convert\main.py", line 176, in convert_xgboost
    return convert(*args, **kwargs)
  File "C:\Users\macke\repos\onnx_sample\env\lib\site-packages\onnxmltools\convert\xgboost\convert.py", line 39, in convert
    model = WrappedBooster(model)
[...]

PR #506, opened 2021-10-05
[WIP] Fix conversion of XGBoost model after being restored

xadupre:
Investigate issue #499.


BowenBao:
Hi @xadupre is the PR ready to be merged?

xadupre:
No, the bug is still here. When the model is unpickled, xgboost does not restore exactly the same object and the converter does not find one attribute it is using.

BowenBao:
> No, the bug is still here. When the model is unpickled, xgboost does not restore exactly the same object and the converter does not find one attribute it is using.

Thanks, I can help take a look.

Issue #475, opened 2021-06-29
ValueError: No proper operator name found for '<class 'pyspark.ml.pipeline.Pipeline'>'


prashanthharshangi:
I am trying to convert a save pyspark pipeline consisting of simple assembler and linear regression model and I ran into 

ValueError: No proper operator name found for '<class 'pyspark.ml.pipeline.Pipeline'>'

Stacktrace below:
```
  File "/Users/prashanth.harshangi/opt/anaconda3/envs/mlpipe/lib/python3.7/site-packages/onnxmltools/convert/main.py", line 77, in convert_sparkml
    custom_conversion_functions, custom_shape_calculators, spark_session)
  File "/Users/prashanth.harshangi/opt/anaconda3/envs/mlpipe/lib/python3.7/site-packages/onnxmltools/convert/sparkml/convert.py", line 69, in convert
    topology = parse_sparkml(spark_session, model, initial_types, target_opset, custom_conversion_functions, custom_shape_calculators)
  File "/Users/prashanth.harshangi/opt/anaconda3/envs/mlpipe/lib/python3.7/site-packages/onnxmltools/convert/sparkml/_parse.py", line 122, in parse_sparkml
    _parse_sparkml(spark, scope, model, inputs, output_dict)
  File "/Users/prashanth.harshangi/opt/anaconda3/envs/mlpipe/lib/python3.7/site-packages/onnxmltools/convert/sparkml/_parse.py", line 90, in _parse_sparkml
    return _parse_sparkml_simple_model(spark, scope, model, global_inputs, output_dict)
  File "/Users/prashanth.harshangi/opt/anaconda3/envs/mlpipe/lib/python3.7/site-packages/onnxmltools/convert/sparkml/_parse.py", line 51, in _parse_sparkml_simple_model
    this_operator = scope.declare_local_operator(get_sparkml_operator_name(type(model)), model)
  File "/Users/prashanth.harshangi/opt/anaconda3/envs/mlpipe/lib/python3.7/site-packages/onnxmltools/convert/sparkml/ops_names.py", line 90, in get_sparkml_operator_name
    raise ValueError("No proper operator name found for '%s'" % model_type)
```

sebimarkgraf:
You are trying to export a unfitted Pipeline.
If you fit the Pipeline it becomes a PipelineModel and can be exported using the onnxmltools.

This is not an issue with the onnxmltools but could be included as helpful warning.

PR #569, opened 2022-06-24
fix: showing better error message when a model of wrong type is being passed.

memoryz:
Signed-off-by: Jason Wang <jasowang@microsoft.com>


memoryz:
For #475.

memoryz:
@xadupre ready for review and merge.

PR #603, opened 2023-01-23
Update utils.py

tittunellimoottil-signifyd:
sparktype is being returned with the parentheses, leading to errors like "Cannot map this type to Onnx types: DoubleType()"


xadupre:
Fixed with #607.

PR #607, opened 2023-02-23
fix: getTensorTypeFromSpark fails for Spark 3.3.0+

memoryz:
The getTensorTypeFromSpark method fails for Spark 3.3.0+. This [PR](https://github.com/apache/spark/pull/34320) added "()" for  `pyspark.sql.types.DataType` string representation.


memoryz:
Just found out this replicate #603

PR #608, opened 2023-02-23
fix: Spark Imputer conversion with multiple input cols

### LINKED ISSUE #596: [Bug] Can't use sparkml imputer for more than 1 variable
I was trying to convert a sparkml pipeline that has an imputer. Got an error and tried to run  _imputer_test_multi  from  TestSparkmlImputer. But keep getting the error.
The error is not limited to the test and i can't run mi pipeline on Onnx.

```
from onnxmltools.convert.sparkml.ops_names import sparkml_operator_name_map
import numpy
import onnxruntime 
from pyspark.ml.feature import Imputer
from onnxconverter_common.onnx_ex import DEFAULT_OPSET_NUMBER
from onnxmltools import convert_sparkml
from onnxmltools.convert.common.data_types import FloatTensorType

data = spark.createDataFrame([
    (1.0, float("nan")),
    (2.0, float("nan")),
    (float("nan"), 3.0),
    (4.0, 4.0),
    (5.0, 5.0)
], ["a", "b"])
imputer = Imputer(inputCols=["a", "b"], outputCols=["out_a", "out_b"])
model = imputer.fit(data)

# the input name should match the inputCols above
model_onnx = convert_sparkml(model, 'Sparkml Imputer Multi Input', [
    ('a', FloatTensorType([None, 1])),
    ('b', FloatTensorType([None, 1]))])

with open("model2.onnx", "wb") as f:
    f.write(model_onnx.SerializeToString())


ss = onnxruntime.InferenceSession("model2.onnx")

input_data = {}
data_np = data.toPandas().values.astype(numpy.float32)
data_np = {'a': data_np[:, :1], 'b': data_np[:, 1:]}

ss.run(None, data_np)   

2022-11-15 10:58:37.834941613 [E:onnxruntime:, sequential_executor.cc:369 Execute] Non-zero status code returned while running Split node. Name:'Split' Status Message: Cannot split using values in 'split' attribute. Axis=1 Input shape={5,2} NumOutputs=2 Num entries in 'split' (must equal number of outputs) was 1 Sum of sizes in 'split' (must equal size of selected axis) was 1
---------------------------------------------------------------------------
Fail                                      Traceback (most recent call last)
/tmp/ipykernel_118391/346882267.py in <module>
      3 data_np = {'a': data_np[:, :1], 'b': data_np[:, 1:]}
      4 
----> 5 ss.run(None, data_np)

~/venv/lib/python3.7/site-packages/onnxruntime/capi/onnxruntime_inference_collection.py in run(self, output_names, input_feed, run_options)
    198             output_names = [output.name for output in self._outputs_meta]
    199         try:
--> 200             return self._sess.run(output_names, input_feed, run_options)
    201         except C.EPFail as err:
    202             if self._enable_fallback:

Fail: [ONNXRuntimeError] : 1 : FAIL : Non-zero status code returned while running Split node. Name:'Split' Status Message: Cannot split using values in 'split' attribute. Axis=1 Input shape={5,2} NumOutputs=2 Num entries in 'split' (must equal number of outputs) was 1 Sum of sizes in 'split' (must equal size of selected axis) was 1
```

But i think i found a solution.
If i change the convert_inputer to this:
```
def convert_imputer(scope, operator, container):
    op = operator.raw_operator

    op_type = 'Imputer'
    
    for i,_ in enumerate(operator.input_full_names):

        attrs = {}
        input_type = operator.inputs[0].type
        surrogates = op.surrogateDF.toPandas().values[0].tolist()
        value = op.getOrDefault('missingValue')
        print(surrogates)
        if isinstance(input_type, FloatTensorType):
            attrs['imputed_value_floats'] = [surrogates[i]]
            attrs['replaced_value_float'] = value
        elif isinstance(input_type, Int64TensorType):
            attrs['imputed_value_int64s'] = [[int(x) for x in surrogates][i]]
            attrs['replaced_value_int64'] = int(value)
        else:
            raise RuntimeError("Invalid input type: " + input_type)

        container.add_node(op_type, operator.input_full_names[i], operator.output_full_names[i],
                            name=scope.get_unique_operator_name('Imputer'),
                           op_domain='ai.onnx.ml',
                           **attrs)

```
And now I can use the imputer.
The problem is that, now, the tests fails because it expects only one array of two columns:
array([[1., 4.],
       [2., 4.],
       [3., 3.],
       [4., 4.],
       [5., 5.]], dtype=float32)

but the output is now two arrays of 1 column each:
[array([[1.],
       [2.],
       [3.],
       [4.],
       [5.]], dtype=float32), array([[4.],
       [4.],
       [3.],
       [4.],
       [5.]], dtype=float32)].

Im not sure if that is a problem. The graph of the original implementation is this:
Screenshot from 2022-11-15 12-16-12

and the new graph is this:
image

Im seeing the same input and same output. 
So, is it okey if i change the test and make a pr?

what do you think?

memoryz:
Fixes #596.

### DIFF
--- a/onnxmltools/convert/sparkml/operator_converters/imputer.py
+++ b/onnxmltools/convert/sparkml/operator_converters/imputer.py
@@ -5,17 +5,19 @@ import copy
 from ...common.data_types import Int64TensorType, FloatTensorType
 from ...common.utils import check_input_and_output_numbers, check_input_and_output_types
 from ...common._registration import register_converter, register_shape_calculator
+from ...common._topology import Operator, Scope
+from pyspark.ml.feature import ImputerModel
+from typing import List
 
-
-def convert_imputer(scope, operator, container):
-    op = operator.raw_operator
-
+def convert_imputer(scope: Scope, operator: Operator, container):
+    op: ImputerModel = operator.raw_operator
     op_type = 'Imputer'
     name = scope.get_unique_operator_name(op_type)
     attrs = {'name': name}
     input_type = operator.inputs[0].type
     surrogates = op.surrogateDF.toPandas().values[0].tolist()
     value = op.getOrDefault('missingValue')
+    
     if isinstance(input_type, FloatTensorType):
         attrs['imputed_value_floats'] = surrogates
         attrs['replaced_value_float'] = value
@@ -37,12 +39,12 @@ def convert_imputer(scope, operator, container):
                            name=scope.get_unique_operator_name('Split'),
                            op_version=2,
                            axis=1,
-                           split=range(1, len(operator.output_full_names)))
+                           split=[1] * len(operator.output_full_names))
     else:
         container.add_node(op_type, operator.inputs[0].full_name, operator.output_full_names[0],
                            op_domain='ai.onnx.ml',
                            **attrs)
-
+        
 
 register_converter('pyspark.ml.feature.ImputerModel', convert_imputer)
 
--- a/onnxmltools/convert/sparkml/operator_converters/imputer.py
+++ b/onnxmltools/convert/sparkml/operator_converters/imputer.py
@@ -44,7 +44,6 @@ def convert_imputer(scope: Scope, operator: Operator, container):
         container.add_node(op_type, operator.inputs[0].full_name, operator.output_full_names[0],
                            op_domain='ai.onnx.ml',
                            **attrs)
-        
 
 register_converter('pyspark.ml.feature.ImputerModel', convert_imputer)
 
--- a/onnxmltools/convert/sparkml/ops_input_output.py
+++ b/onnxmltools/convert/sparkml/ops_input_output.py
@@ -122,8 +122,12 @@ def build_io_name_map():
             lambda model: [model.getOrDefault("predictionCol")],
         ),
         "pyspark.ml.feature.ImputerModel": (
-            lambda model: model.getOrDefault("inputCols"),
-            lambda model: model.getOrDefault("outputCols"),
+            lambda model: model.getOrDefault("inputCols")
+            if model.isSet("inputCols")
+            else [model.getOrDefault("inputCol")],
+            lambda model: model.getOrDefault("outputCols")
+            if model.isSet("outputCols")
+            else [model.getOrDefault("outputCol")],
         ),
         "pyspark.ml.feature.MaxAbsScalerModel": (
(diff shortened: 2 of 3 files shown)


onnxmltools/convert/sparkml/operator_converters/imputer.py
```diff
@@ -37,12 +39,12 @@ def convert_imputer(scope, operator, container):
...
                            op_version=2,
                            axis=1,
-                           split=range(1, len(operator.output_full_names)))
+                           split=[1] * len(operator.output_full_names))
     else:
         container.add_node(op_type, operator.inputs[0].full_name, operator.output_full_names[0],
                            op_domain='ai.onnx.ml',
                            **attrs)
-
+
```
xadupre, line 47: extra spaces

onnxmltools/convert/sparkml/operator_converters/imputer.py
```diff
@@ -37,12 +39,12 @@ def convert_imputer(scope, operator, container):
                            name=scope.get_unique_operator_name('Split'),
                            op_version=2,
                            axis=1,
-                           split=range(1, len(operator.output_full_names)))
+                           split=[1] * len(operator.output_full_names))
```
xadupre, line 42: I assume this piece of code was never tested.

memoryz: Right. With the original code, the unit test fails with this error:
> [E:onnxruntime:, sequential_executor.cc:339 onnxruntime::SequentialExecutor::Execute] Non-zero status code returned while running Split node. Name:'Split' Status Message: Cannot split using values in 'split' attribute. Axis=1 Input shape={5,2} NumOutputs=2 Num entries in 'split' (must equal number of outputs) was 1 Sum of sizes in 'split' (must equal size of selected axis) was 1
And the Split node looks like this:
image
Notice that the split attribute is set to `[1]`.
Based on the [documentation](https://github.com/onnx/onnx/blob/main/docs/Changelog.md#Split-2), the "split" attribute should be set to list of ints for length of each output, so it should be [1, 1] in this case:
image

onnxmltools/convert/sparkml/operator_converters/imputer.py
memoryz: Removed.

memoryz:
@xadupre please merge.

xadupre (approved):

Issue #615, opened 2023-03-20
sparkml utils.py getTensorTypeFromSpark for batch inputs; buildInputDictSimple excludes StringType()


mattoh91:
1. getTensorTypeFromSpark() returns tensor types of shape [1,1]: This should be updated to [None,1] to accommodate batch data input.
2. buildInputDictSimple() only accounts for 'StringType' string dataType: This should also include for 'StringType()'

mattoh91:
Submitted PR to fix issues mentioned above.

PR #616, opened 2023-03-20
Fix for Issue #615

mattoh91:
Updated the following to resolve issue #615 :
1. getTensorTypeFromSpark() now returns tensor types of shape [None,1] to take in batch inputs
2. buildInputDictSimple() now works with string features of dataType 'StringType()'


xadupre:
Replaced by #638.

PR #681, opened 2024-02-14
Handle issue with binary classifier setting output to [N,1] vs [N,2]

casassg:
There's an issue on binarry classifier when passing a booster object.

WrappedBooster will set num_class to 1 as trees//ntrees == 1 https://github.com/casassg/onnxmltools/blob/e7a2ef088fb336800d34bffa628abac79f1ecc14/onnxmltools/convert/xgboost/_parse.py#L100

This makes it such that when we try to calculate the ncl for the classifier we end up setting an output of N, 1 vs N,2 which is what ONNX will actually calculate. These produce errors on logs in onnx runtime.

Given binary:logistic will always produce [N,2] changing the order in the if clause should remove the error.

I can add a test to repro but wanted to send PR first for feedback.

Also removed a double assignation which I'm not certain why it was there.


casassg:
@xadupre any chance to do a release w this change?

PR #683, opened 2024-02-23
Fix pkg name of onnxconverter_common

xkszltl:
The pkg is dashed while the lib is underscored

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 numpy
 onnx
-onnxconverter_common
+onnxconverter-common


xadupre (approved):

XiangRongLin:
Am I seeing correctly that although the fix is merged, no release was created yet? Because I still encounter this issue when upgrading to 1.12.0

xkszltl:
You may want to push project owner for a patch release.

xkszltl:
1.12.0 is from last year.

Issue #712, opened 2025-04-18
Integrate with ONNX 1.18.0 release branch


ramkrishna2910:
We are releasing ONNX 1.8.0. A release branch is created (https://github.com/onnx/onnx/tree/rel-1.18.0). Release candidates are also available from TestPyPI: `pip install -i https://test.pypi.org/simple/ --pre onnx`

Updates to op-list and other key updates can be found here: https://github.com/onnx/onnx/wiki/Logistics-for-ONNX-Release-1.18.0

It is important to integrate ONNX release branch ASAP so that any issues and incompatibilities can be detected and resolved before the ONNX release.

In case a bug in ONNX is detected during integration of ONNX 1.18.0, please open a [ONNX Bug Report](https://github.com/onnx/onnx/issues/new?assignees=&labels=bug&projects=&template=bug.md&title=) and tag ONNX Release Manager @ramkrishna2910 so that the bug is fixed in the ONNX release branch.

ad-astra-video:
Getting error `ImportError: cannot import name 'split_complex_to_pairs' from 'onnx.helper'` which looks like deprecated a number of functions. This specific one was moved to `_split_complex_to_pairs`.

https://github.com/onnx/onnx/commit/bcecb8c115274cb86d7599794e5c50d62d21f4ef#diff-9cd90c72fd9222eb7ca56d100fad24abbb09bfd6a6a77d3fdf929d9a4ec3bd5f

PR #713, opened 2025-05-14
Remove import of split_complex_to_pairs and unused functions

justinchuby:
Compatibility with onnx 1.18

Fix https://github.com/onnx/onnxmltools/issues/712


xadupre:
Replaces #713.

PR #714, opened 2025-05-15
Remove import of split_complex_to_pairs and unused functions

xadupre:
Replaces #713


xadupre:
I'm updated the process to publish a package for sklearn-onnx. Once it is done, I'll do the same for onnxmltools. It should be soon.

davidmezzetti:
Hello @xadupre just wanted to see if this release was still planned. I was going to wait on this before releasing the next version of txtai.

Issue #732, opened 2025-10-23
Support xgboost v3


AsherWright:
## Overview

xgboost v3 changed `base_score`, such that we hit `ValueError: could not convert string to float: '[4.6E-1]'`

I believe we'll have to handle it in [xgboost/common.py](https://github.com/onnx/onnxmltools/blob/main/onnxmltools/convert/xgboost/common.py#L31), e.g.

From
```python
if "base_score" in config["learner"]["learner_model_param"]:
        bs = float(config["learner"]["learner_model_param"]["base_score"])
        # xgboost >= 2.0
        params["base_score"] = bs
```

To
```python
if "base_score" in config["learner"]["learner_model_param"]:
        bs = config["learner"]["learner_model_param"]["base_score"]
        if isinstance(bs, str) and bs.startswith("["):
            # xgboost >= 3.0 stores base_score as an array string '[value]'
            bs = float(bs.strip("[]").split(",")[0].strip())
        else:
            # xgboost >= 2.0, < 3.0 stores base_score as a string 'value'
            bs = float(bs)

        params["base_score"] = bs
```

PR #733, opened 2025-10-25
Fix intercept handling of xgboost starting with version 3.1.0

lukaswenzl:
Starting with version 3.1.0 xgboost changed the default behavior of base_score to be a vector instead of a number: 
see https://xgboost.readthedocs.io/en/stable/parameter.html

> Changed in version 3.1.0: XGBoost is updated to use vector-valued intercept by default.

This breaks the conversion script to ONNX, and most xgboost tests of this repo fail when using the latest version of xgboost. 

This issue was also raised in https://github.com/onnx/onnxmltools/issues/732

In this pr I am proposing a fix:
* add support for reading base_score as a number or a list of numbers
* If the list contains only one number then convert to a number and proceed as before
* If there are multiple numbers, this indicates we have a multiclass model with different intercepts for each class. We can store the whole list and map it to base_values, which can already handle distinct intercept values for each class. 

With these changes all tests pass again for the latest version of xgboost

### DIFF
--- a/onnxmltools/convert/xgboost/_parse.py
+++ b/onnxmltools/convert/xgboost/_parse.py
@@ -121,9 +121,16 @@ def _get_attributes(booster):
             kwargs["objective"] = "binary:logistic"
 
     if "base_score" not in kwargs:
-        kwargs["base_score"] = 0.5
+        kwargs["base_score"] = [0.5]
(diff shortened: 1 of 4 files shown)


AsherWright:
Added support for multiclass problems here: https://github.com/onnx/onnxmltools/pull/736

